PR #6, opened 2018-11-29
Add margin with a transition to the seek bar

nycex:
Add margin with a transition to the seek bar and accordingly fixes the overlapping. I don't know if you like it, so if not just close.


aetheryx:
discussed in the discord server - it's weird when the bar itself moves position on hover, bad UX

PR #52, opened 2019-01-04
Local track and unloaded track error.

SacreBleuXCVII:
Allows for the local track to be shown, wasn't sure what to set the image to so left it null for now.
The other error was when `playerState` sometimes showed as "unknown" and didn't contain the required data so threw an error, tended to happen when Spotify hadn't loaded any tracks.

### DIFF
--- a/src/Powercord/plugins/pc-spotify/Modal/Modal.jsx
+++ b/src/Powercord/plugins/pc-spotify/Modal/Modal.jsx
@@ -35,7 +35,7 @@ module.exports = class Modal extends React.Component {
   }
 
   updateData (playerState) {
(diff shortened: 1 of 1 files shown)


src/Powercord/plugins/pc-spotify/Modal/Modal.jsx
```diff
@@ -35,12 +35,12 @@ module.exports = class Modal extends React.Component {
   }
 
   updateData (playerState) {
-    if (playerState) {
+    if (playerState && playerState.currently_playing_type != 'unknown') {
```
cyyynthia: !==

cyyynthia (changes requested):

cyyynthia (approved):

PR #54, opened 2019-01-04
dependency updates

davidcralph:
again if this pull request is gross just close pls kthx


aetheryx:
both of those are noted with a caret, so this doesn't make a difference afaik

davidcralph:
looks nicer tho :badboy:

davidcralph:
yeeting the close button i guess

PR #55, opened 2019-01-06
Mock Plugin

melmsie:
I realize this isn't super super super necessary, however I see people mock in aeth's server all the time so I'm sure the demand is there for an easier way to do it.


cyyynthia (approved):
This plugin can be later changed to a kind of extended formatting with more formatting styles, but it's a useful plugin given how annoying is to TyPe ThAt LiKe ThIs

Issue #53, opened 2019-01-04
[ Commands ] Auto completion keeps rendering when going behind the prefix


MrLar:
This causes the auto completion feature to be inconsistent with other auto completions within discord.

For reference: https://shuvi.site/i/307434.mp4

PR #56, opened 2019-01-06
A fix and a feature

melmsie:


melmsie:
Not sure why mock is in this, but ok

Fixes: https://france-is-a-me.me/15a92f.png
Feature: Closes #53

src/Powercord/plugins/pc-commands/injectAutocomplete.js
```diff
@@ -26,6 +26,10 @@ module.exports = async function injectAutocomplete () {
         )
       }),
       renderResults: (...args) => {
+        const textArea = document.getElementsByClassName('pc-textArea');
+        if (textArea[0].selectionStart === 0) {
+          return;
```
aetheryx: this block of code should be in the `matches` function instead, and it should just check for the provided instance state, i.e.
```js
matches: () => startsWith(this.prefix) && this.instance.state.selectionEnd !== 0
```

aetheryx (changes requested):

src/Powercord/plugins/pc-commands/injectAutocomplete.js
melmsie: Convo about this on discord [starts here](https://canary.discordapp.com/channels/299979631715549184/517348857219383308/531479012556537856) for the sake of recording stuff

aetheryx (approved):

PR #67, opened 2019-01-13
Emote Utility Update

21Joakim:
* Changes
  * Changed the command "find" to a less generic name "findemote".

</br>

* New features
  * You can now input multiple emotes in to "findemote"
  * You can now use the "saveemote" command to save emotes to a pre-defined directory
  * You can now use the "cloneemote" command to clone emotes to your own server (The server argument can default to the current guild or a pre-defined guild, check the settings)

</br>

* Added settings for
  * Responding with embeds or not
    * Setting to have a link to the guild in "findemote" when this is disabled
  * Set a save directory for the "saveemote" command
  * Use current guild when cloning emotes
    * Or set a pre-defined guild

### DIFF
--- a/src/Powercord/plugins/pc-emojiUtility/Settings.jsx
+++ b/src/Powercord/plugins/pc-emojiUtility/Settings.jsx
@@ -1,0 +1,87 @@
+const { React } = require('powercord/webpack');
+const { SwitchItem, TextInput } = require('powercord/components/settings');
+const { existsSync, lstatSync } = require('fs');
+
+module.exports = class EmojiUtilitySettings extends React.Component {
+  constructor (props) {
+    super();
+
+    this.settings = props.settings;
+    this.state = {
+      useEmbeds: props.settings.get('useEmbeds', true),
+      displayLink: props.settings.get('displayLink', true),
+      filePath: props.settings.get('filePath', null),
+      
+      isFilePathValid: props.settings.get('filePath') ? existsSync(props.settings.get('filePath')) : true,
+      initialFilePathValue: props.settings.get('filePath')
+    };
+  }
+
+  render () {
+    const settings = this.state;
+
+    const set = (key, value = !settings[key], defaultValue) => {
+      if (!value && defaultValue) {
+        value = defaultValue;
+      }
+
+      this.settings.set(key, value);
+      this.setState({
+        [key]: value
+      });
+    };
+
+    return (
+      <div>
+        <SwitchItem
+          note='Whether Emote Utility should return responses in embeds.'
+          value={settings.useEmbeds}
+          onChange={() => set('useEmbeds')}
+        >
+          Use embeds
+        </SwitchItem>
+
+        {
+          (() => {
+            if(settings.useEmbeds) {
+              return;
+            }
+
+            return (
+              <SwitchItem
+                note='Whether the message for the findemote command should contain the link to the guild the emote is in.'
+                value={settings.displayLink}
+                onChange={() => set('displayLink')}
+              >
+                Display link
+              </SwitchItem>
+            )
+          })()
+        }
+
+        <TextInput
+          note='The file path emotes should be saved to when using the saveemote command'
+          defaultValue={settings.filePath}
+          style={!this.state.isFilePathValid ? {borderColor: 'red'} : {}}
+          onChange={(value) => {
+            if(value.length === 0 || (existsSync(value) && lstatSync(value).isDirectory())) {
+              this.setState({
+                isFilePathValid: true
+              });
+
+              set('filePath', value.length === 0 ? null : value);
+            }else{
+              this.setState({
+                isFilePathValid: false
+              });
+
+              set('filePath', this.state.initialFilePathValue);
+            }
+          }}
+        >
+          File path
+        </TextInput>
+      </div>
+    );
+  }
+};
--- a/src/Powercord/plugins/pc-emojiUtility/Settings.jsx
+++ b/src/Powercord/plugins/pc-emojiUtility/Settings.jsx
@@ -1,6 +1,8 @@
-const { React } = require('powercord/webpack');
+const { React, getModule } = require('powercord/webpack');
 const { SwitchItem, TextInput } = require('powercord/components/settings');
 const { existsSync, lstatSync } = require('fs');
+
+const { getGuild } = getModule([ 'getGuild' ]);
 
 module.exports = class EmojiUtilitySettings extends React.Component {
   constructor (props) {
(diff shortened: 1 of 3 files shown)


src/Powercord/plugins/pc-emojiUtility/index.js
```diff
@@ -1,42 +1,173 @@
 const Plugin = require('powercord/Plugin');
-const { getModule } = require('powercord/webpack');
+const { getModule, channels: { getSelectedChannelState, getChannelId }, constants: { Routes, APP_URL_PREFIX }, React, messages: { createBotMessage, receiveMessage } } = require('powercord/webpack');
+
+const { existsSync, createWriteStream } = require('fs');
+const { get } = require('https');
```
cyyynthia: Use `powercord/http` module

src/Powercord/plugins/pc-emojiUtility/index.js
```diff
@@ -1,42 +1,173 @@
...
   }
 
+  getGuildUrl (guild) {
+    let selectedChannel = getSelectedChannelState()[guild.id];
+    if (!selectedChannel) {
+      /* I am not sure if it can get here but just to be sure */
+      selectedChannel = guild.systemChannelId;
+    }
+
+    return APP_URL_PREFIX + Routes.CHANNEL(guild.id, selectedChannel); // eslint-disable-line new-cap
```
cyyynthia: You should use `Routes.GUILD` like the experimental `pc-guilds` plugin do

src/Powercord/plugins/pc-emojiUtility/index.js
```diff
@@ -1,42 +1,173 @@
...
+  download (url) {
+    return new Promise((resolve, reject) => {
+      get(url)
+        .on('response', (response) => resolve(response))
+        .on('error', (error) => reject(error));
+    });
+  }
+
+  write (path, stream) {
+    return new Promise((resolve, reject) => {
```
cyyynthia: you can use `writeFile` from fs.promises, so it's just `writeFile(path, contents);`

src/Powercord/plugins/pc-emojiUtility/index.js
```diff
@@ -50,20 +181,218 @@ module.exports = class EmojiUtility extends Plugin {
...
+                ? {
+                  type: 'rich',
+                  description: 'Please provide an emote',
+                  color: 16711680
+                }
+                : 'Please provide an emote'
+            };
+          }
+
+          const emojis = [ ...new Set(Object.values(emojiStore.getGuilds()).flatMap(r => r.emojis)) ];
```
cyyynthia: looks like duplicated code, should be moved to a function

src/Powercord/plugins/pc-emojiUtility/index.js
```diff
@@ -50,20 +181,218 @@ module.exports = class EmojiUtility extends Plugin {
...
+              : `Could not find any emotes containing **${argument}**`
           };
         }
       );
+
+    powercord
+      .pluginManager
+      .get('pc-commands')
+      .register(
+        'saveemote',
```
cyyynthia, line 249: I'm not really sure if this is really something needed. Discord already allows you to open an emote in browser and download it. Imo, it would be util if this was able to re-upload emojis automatically to a server instead of letting ppl re-upload them

src/Powercord/plugins/pc-emojiUtility/index.js
```diff
@@ -50,20 +181,218 @@ module.exports = class EmojiUtility extends Plugin {
...
+                color: 16711680
+              }
+              : `**${notFoundEmojis.length}** of the provided arguments ${notFoundEmojis.length === 1 ? 'is not a custom emote' : 'are not custom emotes'}`
+            );
+          }
+
+          if (foundEmojis.length < 5) {
+            for (const emoji of foundEmojis) {
+              try {
+                await this.write(filePath + emoji.name + extname(parse(emoji.url).pathname), await this.download(emoji.url));
```
cyyynthia: path.resolve

cyyynthia (changes requested):
I reviewed that kinda quickly but there is some stuff to update

src/Powercord/plugins/pc-emojiUtility/index.js
21Joakim: Going to make it context later for now it can be a command

image

src/Powercord/plugins/pc-emojiUtility/Settings.jsx
```diff
@@ -0,0 +1,144 @@
...
+      <div>
+        <SwitchItem
+          note='Whether Emote Utility should return responses in embeds.'
+          value={settings.useEmbeds}
+          onChange={() => set('useEmbeds')}
+        >
+          Use embeds
+        </SwitchItem>
+
+        {
```
cyyynthia:
```jsx
{settings.useEmbeds && <SwitchItem .../>}
```

src/Powercord/plugins/pc-emojiUtility/index.js
```diff
@@ -1,42 +1,190 @@
...
+      input = input.toLowerCase();
+
+      guild = Object.values(getGuilds()).find(g => g.name.toLowerCase().includes(input));
+    }
+
+    return guild;
+  }
+
+  findEmojisForCommand (args) {
+    args = [ ...new Set(args) ];
```
cyyynthia, line 66: why?

cyyynthia (changes requested):
I read that quicly, but I think non-nitro users will not really be able to use this plugin, especially the cloneemote thing. Also, I find this too command-based and some stuff (like the emote stealing feature) can be moved to the context menu. Even better, add a "get info" when you right click an emote, and then pop a dialog with the emote, the name, the emote id, the guild where it's from, a reupload button, a download button (and maybe instead of having that in settings selecting it here directly), etc etc

Tho that's stuff can be done in a v2 version of that plugin :)

src/Powercord/plugins/pc-emojiUtility/index.js
aetheryx: makes the items unique afaik

cyyynthia: makes sense

21Joakim:
What I planned to do was just make command versions of everything first and then, later on, make context menu stuff. 

And the "cloneemote" command can easily be used by someone that doesn't have nitro because it has a second argument that is the guild you want to clone it to, which takes an ID or the guild name, so they would just have to go to the guild the emote is in and use ".cloneemote :emote: Aetheryx' Hub" and it would be cloned.

cyyynthia:
and for name conflicts? The best way is to have the ID of the emote, which is easy for someone who have Nitro, and harder for someone who don't have Nitro

21Joakim:
I think you misunderstood what I said, you would still have to send the emote but you can specify which guild it would go in to. So instead of being able to use the command from any guild you would use it in the guild the emote is from.

cyyynthia:
Then re-uploading emote to server is annoying since it requires to set the guild to upload each time
For now let's consider it works well enough to be merged

cyyynthia (approved):

PR #87, opened 2019-02-01
Fix spelling mistake

davidcralph:
bowser pls


cyyynthia:
ok next plugin IDs are gonna be in french

davidcralph:
i agree

PR #95, opened 2019-02-07
Justice to Clyde!

21Joakim:
Help a local Clyde out

Before:
discordcanary_cqnn2ton5t

After:
discordcanary_tq0quypfqp


aetheryx:
Could you elaborate..? I'm not sure why we would change this from Discord's default

21Joakim:
Based on what people thought in the official powercord server (Powercord should not change any default CSS as well as this could potetionally mess up themes) I will close this issue.

PR #96, opened 2019-02-07
[Commands] Don't wait for disabled plugins

21Joakim:
Having a disabled plugin would cause the autocomplete to not load until you re-enabled it


cyyynthia (changes requested):
Add pc-classNameNormalizer to dependencies

21Joakim:
Done 👌

davidcralph:
very cool

cyyynthia (approved):

PR #140, opened 2019-03-17
Repair 'Announcements' plug-in

GriefMoDz:
### Plugin Changelog (v0.1.1) :: Announcements
index.js:
* Add notice types to current announcements (first-welcome & pewdiepie).
* Modify message content, button text and `onClick()` method for `pc-pewdiepie`.
* Announcements that are set to `alwaysDisplay` no longer write to the "dismissed" array found in settings.
* Update query selectors to use `.pc-wrapper` class instead of `.pc-guildWrapper` as it is now obsolete.

Notice.jsx:
* Add an additional notice type based around `noticeInfo` that changes the CTA button text to black on hover.

### DIFF
--- a/src/Powercord/plugins/pc-announcements/Notice.jsx
+++ b/src/Powercord/plugins/pc-announcements/Notice.jsx
@@ -18,27 +18,24 @@ const Notice = class Notice extends React.Component {
   async componentDidMount () {
     if (!classesStore) {
       const classes = await getModule([ 'noticeBrand' ]);
-      Promise.all([ classes ])
-        .then(value => {
-          this.setState({
-            types: {
-              blurple: value[0].noticeBrand,
-              red: value[0].noticeDanger,
-              orange: value[0].noticeDefault,
-              facebook: value[0].noticeFacebook,
-              blue: value[0].noticeInfo,
-              dark: value[0].noticePremium,
-              blurple_gradient: value[0].noticePremiumGrandfathered,
-              spotify: value[0].noticeSpotify,
-              purple: value[0].noticeStreamerMode,
-              green: value[0].noticeSuccess,
-              survey: value[0].noticeSurvey
-            },
-            button: value[0].button,
-            dismiss: value[0].dismiss
-          });
-          classesStore = this.state;
-        });
+      this.setState({
+        types: {
+          blurple: classes.noticeBrand,
+          red: classes.noticeDanger,
+          orange: classes.noticeDefault,
+          facebook: classes.noticeFacebook,
+          blue: classes.noticeInfo,
+          dark: classes.noticePremium,
+          blurple_gradient: classes.noticePremiumGrandfathered,
+          spotify: classes.noticeSpotify,
+          purple: classes.noticeStreamerMode,
+          green: classes.noticeSuccess,
+          survey: classes.noticeSurvey
+        },
+        button: classes.button,
+        dismiss: classes.dismiss
+      });
+      classesStore = this.state;
     }
   }
 
--- a/src/Powercord/plugins/pc-announcements/Notice.jsx
+++ b/src/Powercord/plugins/pc-announcements/Notice.jsx
@@ -1,4 +1,4 @@
-const { React, getModule, getModuleByDisplayName } = require('powercord/webpack');
+const { React, getModule, getComponentByDisplayName } = require('powercord/webpack');
 
 const Clickable = getComponentByDisplayName('Clickable');
 
@@ -43,7 +43,7 @@ const Notice = class Notice extends React.Component {
     const { notice, onClose } = this.props;
     const { types, button, dismiss } = this.state;
 
-    return <div className={`powercord-notice ${(types[notice.type] || types.blurple)}`}>
+    return <div className={`powercord-notice ${(types[notice.type] || types.BLURPLE)}`}>
       {notice.message}
       <Clickable className={dismiss} onClick={() => onClose()}/>
       {notice.button &&
--- a/src/Powercord/plugins/pc-announcements/Notice.jsx
+++ b/src/Powercord/plugins/pc-announcements/Notice.jsx
@@ -1,6 +1,6 @@
 const { React, getModule, getModuleByDisplayName } = require('powercord/webpack');
 
-const Clickable = getModuleByDisplayName('Clickable');
+const Clickable = getComponentByDisplayName('Clickable');
 
 let classesStore = null;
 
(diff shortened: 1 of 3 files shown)


src/Powercord/plugins/pc-announcements/index.js
```diff
@@ -7,8 +7,6 @@ const { inject, uninject } = require('powercord/injector');
 const { React, ReactDOM, getModule } = require('powercord/webpack');
 const { DISCORD_INVITE, GUILD_ID } = require('powercord/constants');
 
-const Notice = require('./Notice');
```
cyyynthia: why removing this? It's used in _renderNotice

src/Powercord/plugins/pc-announcements/index.js
```diff
@@ -50,15 +53,18 @@ module.exports = class Announcements extends Plugin {
   }
 
   sendNotice (notice, alwaysDisplay) {
-    if (!this.notices.find(n => n.id === notice.id) && (alwaysDisplay || !this.settings.get('dismissed', []).includes(notice.id))) {
+    notice.alwaysDisplay = alwaysDisplay || false; // append a new key to the corresponding notice; we'll use this to grab the value of alwaysDisplay outside of this method
```
cyyynthia: Then you can remove the alwaysDisplay parameter and let developers put the parameter in the notice itself

src/Powercord/plugins/pc-announcements/index.js
```diff
@@ -50,15 +53,18 @@ module.exports = class Announcements extends Plugin {
...
+    if (!this.notices.find(n => n.id === notice.id) && (notice.alwaysDisplay || !this.settings.get('dismissed', []).includes(notice.id))) {
       this.notices.push(notice);
       this._renderNotice();
     }
   }
 
   closeNotice (noticeId) {
     this.notices = this.notices.filter(n => n.id !== noticeId);
-    this.settings.set('dismissed', [ ...this.settings.get('dismissed', []), noticeId ]);
+    if (this.notices.find(n => n.alwaysDisplay)) { // make sure that we're only adding notices found without alwaysDisplay to the 'dismissed' array
```
cyyynthia: this will prevent *any* notice to be added to the dismissed store if any announcement have alwaysDisplay on, even if the notice don't have it. You need to grab the notice using its ID (before we remove it from the store), and then check for the value.

src/Powercord/plugins/pc-announcements/index.js
```diff
@@ -73,16 +79,17 @@ module.exports = class Announcements extends Plugin {
   }
 
   _renderNotice () {
-    if (document.querySelector('.pc-guildsWrapper + .pc-flex > .pc-flexChild .pc-notice')) {
+    const Notice = require('./Notice');
```
cyyynthia: why put the require here?

cyyynthia (changes requested):

src/Powercord/plugins/pc-announcements/index.js
```diff
@@ -31,16 +30,21 @@ module.exports = class Announcements extends Plugin {
               getModule([ 'selectGuild' ]).selectGuild(GUILD_ID);
             });
           }
-        }
-      }, true);
+        },
+        alwaysDisplay: true,
```
cyyynthia: eslint won't like this `,` 👀
```suggestion
        alwaysDisplay: true
```

cyyynthia (changes requested):

src/Powercord/plugins/pc-announcements/index.js
GriefMoDz: > why removing this? It's used in _renderNotice

No need to be alarmed. I'm aware that this constant variable is used in `_renderNotice` thus I've only went ahead and ~~re~~moved it to the location of said method (reasonings have been stated below).

src/Powercord/plugins/pc-announcements/index.js
GriefMoDz: > why put the require here?

To answer **why** I've chosen to move the require here compared to on `Line 10` is mainly due to the fact that the `Notice.TYPES` store is called too early (way before the notice classes from `getModule([ 'noticeBrand' ])` are even initialized).

That said, with the constant variable staying in it's original spot (on line 10) you'll find that the classes for announcements/notices can't be retrieved, causing them to be `undefined` and ultimately preventing styling (CSS) to both the `<div>` and its corresponding elements (`<button>` and `<Clickable>`).

cyyynthia: I see why the require was moved, but since _renderNotice can still be called early it can still cause race conditions in case of slow boots. In cases where the module is not initialized, Powercord will return a promise and it breaks as it's not what you'd expect.

A workarround can be to not pass classes directly in Notice.TYPES but just give constants (but keep the object for autocomplete/compatibility purposes) (`BLURPLE: 'BLURPLE'`) and mount classes in the `componentDidMount` with await and all promise stuff. Just use setState to store classes and make the component re-render, and boom.

What you can also do is save it in a variable and initialize store with it immediately, so if the module already have been fetched, it won't fetch it again.

GriefMoDz: Marking this conversation as **Resolved** - workaround can be found in the latest commit. Let me know if anything else needs to be changed.

src/Powercord/plugins/pc-announcements/Notice.jsx
```diff
@@ -1,32 +1,72 @@
...
+      button: '',
+      dismiss: ''
+    };
+  }
+
+  async componentDidMount () {
+    if (!classesStore) {
+      const classes = await getModule([ 'noticeBrand' ]);
+      Promise.all([ classes ])
+        .then(value => {
```
aetheryx: This Promise.all() call seems unnecessary to me. `classes` is already an awaited variable, i.e. a resolved promise. You should be able to kill `Promise.all([ classes ]).then(value => { ... })` and replace all instances of `value[0]` with `classes`.

aetheryx (changes requested):

src/Powercord/plugins/pc-announcements/Notice.jsx
GriefMoDz: That's a good call! Honestly not sure why I chose to use the `Promise.all()` call when there's only ever one key returned from the `noticeBrand` module. Anyway, I've gone ahead and added these requested changes to my latest commit.

cyyynthia (approved):
I pushed some changes to the PR (as discussed in Discord), now waiting for aeth's approval 👀

aetheryx (approved):

PR #144, opened 2019-03-21
[Spotify] Add filler image

### LINKED ISSUE #60: [ Suggestion ] [ Spotify ] Add some kind of filler image when playing local songs
The spotify modal looks kind of weird when you play a local file due to the huge empty space created by the fact that there is no thumbnail.

davidcralph:
Closes #60

### DIFF
--- a/src/Powercord/plugins/pc-spotify/Modal/Modal.jsx
+++ b/src/Powercord/plugins/pc-spotify/Modal/Modal.jsx
@@ -52,7 +52,8 @@ module.exports = class Modal extends React.Component {
         progress: 0,
         progressAt: Date.now()
       },
-      displayState: 'hide'
+      displayState: 'hide',
+      spotifyLogo: ''
     };
   }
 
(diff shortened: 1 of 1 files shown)


davidcralph:
works very well yes

src/Powercord/plugins/pc-spotify/Modal/Modal.jsx
```diff
@@ -89,11 +89,12 @@ module.exports = class Modal extends React.Component {
 
   updateData (playerState) {
     if (playerState && playerState.currently_playing_type === 'track') {
-      return this.setState({
+      return powercord.pluginManager.get('pc-spotify').getSpotifyLogo().then(value => {
```
cyyynthia: I'm not a fan of fetching the logo everytime the player state is updated. What you should to is save that in an object property, and re-use it if the img is `null`

cyyynthia (changes requested):

src/Powercord/plugins/pc-spotify/Modal/Modal.jsx
davidcralph: yikes i don't get that my JS knowledge is failing me 2.0

GriefMoDz:
In reference to @Bowser65's requested changes, here's how I would go about in assigning the logo to an object property and also re-use it's value if `currentItem.img` is `null`: [click me!](https://gist.github.com/GriefMoDz/09c06acc89c6b3d52b8e8ea93269006c)

**Note**: As opposed to linking a gist link, I wasn't able to request my own changes due insufficient permissions (at least not on other lines) so I apologise if this comes across as an inconvience.

davidcralph:
Implemented that gist, thanks! 👍

cyyynthia (approved):

aetheryx (approved):

PR #146, opened 2019-03-24
[API Refactor] 1st breaking changes

cyyynthia:
# Summary
 - Renamed `start` and `unload` plugin main functions to `pluginDidLoad` and `pluginWillUnload`;
These names are more similar to React and feel more consistent (imo) given the fact that we are in a React environment
 - Moved settings registering to an API;
 - Make plugin unregister settings automatically in the background;
 - Started adding some logic to shutdown and restart Powercord at runtime;
 - Made the 1st 3 tabs consistent in terms of order (General Settings, Plugins and Updater)

More breaking changes are to come; Commands will soon follow the same logic as settings, and the StyleManager will get moved to a top-level API (like `powercord.pluginManager`)


src/Powercord/apis/settings/index.js
```diff
@@ -0,0 +1,157 @@
...
+    let panelContents;
+    try {
+      panelContents = React.createElement(contents);
+    } catch (e) {
+      this.error('Failed to render settings panel, check if your function returns a valid React component!');
+      panelContents = null;
+    }
+
+    const h2 = React.createElement(getComponentByDisplayName('FormTitle'), { tag: 'h2' }, title);
+    return React.createElement(getComponentByDisplayName('FormSection'), {}, h2, panelContents);
```
aetheryx, line 82: is this safe to be not awaited?

cyyynthia: getComponentByDisplayName is an helper to get an AsyncComponent instance that will await the module by itself. No need to await it

aetheryx (approved):

PR #153, opened 2019-04-06
[Announcements] Fix 'Join Server' button

GriefMoDz:
By swapping out the `selectGuild` module with `flushSelection`, this patch fixes the mishap that appears when navigating away from the 'Activity' tab moments after selecting the 'Join Server' button (shown after successfully injecting Powercord).

Thanks, @ohlookitsderpy for bringing this to my attention otherwise I never would've known about it. 😊


cyyynthia:
If you could also include checking if the user is already in the server and not call acceptInvite if it's the case would be awesome :)

cyyynthia (approved):

PR #155, opened 2019-04-08
API Refactor

cyyynthia:
I'll update that PR when stuff gets pushed, feel free to give some suggestions.
Plugin devs can also use this to know what will break lol

 - [x] Move commands to an API
 - [x] Move keybinds to an API
 - [x] Move StyleManager to a manager
 - [x] Make plugins register a list of permissions in manifest shown when installing
 - [x] Reorganize `powercord/webpack`
      - [x] Make `getModule` and other functions always return a promise
      - [x] Move `getComponentByDisplayName`
 - [ ] Make `pc-updater` more modulable internally to handle updating plugin and themes
 - [x] Add an `optionalDependencies` key in plugin manifest
 - [x] (won't fix) Optimize `_fetchAllModules` in `pc-classNameNormalizer`
 - [x] Create a Flux store for settings


src/Powercord/managers/plugins.js
```diff
@@ -56,34 +48,9 @@ module.exports = class PluginManager {
...
-        deps.push(dep);
-        deps.push(...(await this.resolveDependencies(dep, deps)));
-      }
-    }));
-    return deps.filter((d, p) => deps.indexOf(d) === p);
-  }
-
   resolveDependents (plugin, dept = []) {
-    const dependents = this.getAllPlugins().filter(p => this.getDependenciesSync(p).includes(plugin));
+    const dependents = this.getPlugins().filter(p => this.getDependenciesSync(p).includes(plugin));
```
GriefMoDz, line 53: Change the callback function found under 'filter' to `this.get(p).allDependencies.includes(plugin)` as the `getDependenciesSync(p)` method was removed (since 70c1c05) and is now deprecated.

**Note**: Leaving this line untouched prevents plug-ins from being disabled (or uninstalled) straight from the 'Plugins' tab of user settings.

src/Powercord/apis/settings/index.js
```diff
@@ -78,27 +85,22 @@ module.exports = class Settings extends API {
...
-    const h2 = React.createElement(getComponentByDisplayName('FormTitle'), { tag: 'h2' }, title);
-    return React.createElement(getComponentByDisplayName('FormSection'), {}, h2, panelContents);
+    const h2 = React.createElement(FormTitle, { tag: 'h2' }, title);
+    return React.createElement(FormSection, {}, h2, panelContents);
   }
 
   // @todo: Discord settings sync
   async upload () {
-    if (!powercord.account || !this.settings['pc-general'].get('settingsSync', false)) {
+    if (!powercord.account || !this.store.settings['pc-general'].get('settingsSync', false)) {
```
GriefMoDz: Change `this.store.settings['pc-general'].get('settingsSync', false)` to `this.store.settings.getSetting('pc-general', 'settingsSync', false)` as the `get` method was removed (since 4bc68e9) and is now deprecated.

**Note**: The same applies to [line 131](https://github.com/powercord-org/powercord/blob/10e86cd5d4cf39e13d3b4bd85f6a6be6753f5827/src/Powercord/apis/settings/index.js#L131) where the old method remains still in use.

src/Powercord/managers/plugins.js
cyyynthia: I'm aware, but keep in mind that the plugin manager have never been here to work, that's why there is a red banner :)

src/Powercord/apis/settings/index.js
cyyynthia: yh aware of this one and fixed locally, just didn't pushed lol

src/Powercord/apis/keybinds.js
```diff
@@ -0,0 +1,75 @@
...
+    const bind = this.keybinds.find(k => k.id !== keybind);
+    if (bind) {
+      this._unregister(bind.keybind, bind.global);
+      this.keybinds = this.keybinds.filter(k => k.id !== keybind);
+    }
+  }
+
+  _register (keybind, func, global) {
+    try {
+      if (global) {
```
aetheryx: don't shadow global pls. -> isGlobal

src/Powercord/plugins/pc-pluginManager/components/Plugin/Container.jsx
```diff
@@ -0,0 +1,41 @@
+const { React } = require('powercord/webpack');
+const {
+  Tooltip, Icons: {
```
aetheryx: disgusting. put icons on a new line

PR #164, opened 2019-04-17
[Spotify] Add lyrics command

davidcralph:
This *could* be part 1 of a two-part pull request, but at the moment React is being a pain in the ass. However, this command is ready to be merged. Tested and confirmed to be working on latest Discord Canary Windows 10.

Let me know if anything needs changing!


cyyynthia:
We've internally discussed about that PR, and we decided we won't accept it.

Major reason is the use of an external API that is neither from Powercord or an official trusted entity. We've updated [guidelines](https://github.com/powercord-community/guidelines) to reflect that statement.

Make it an external plugin (with proper manifest keys) :)

davidcralph:
Understandable

PR #166, opened 2019-04-17
Fix titleBarGames, which was crashing the client

gear4s:
Also fix style of gamebar's last child


gear4s:
plz merge my pr the update popup wont stop popping up

PR #172, opened 2019-04-22
[Emote Utility] Fixed to work with the API-refactor

21Joakim:


davidcralph:
shouldn't this plugin be moved into its own repo now?

src/Powercord/plugins/pc-emojiUtility/components/Settings.jsx
```diff
@@ -2,26 +2,61 @@ const { existsSync, lstatSync } = require('fs');
...
 
-const { getGuild } = getModule([ 'getGuild' ]);
-const { getSortedGuilds } = getModule([ 'getSortedGuilds' ]);
+let getGuild, getSortedGuilds;
 
 module.exports = class EmojiUtilitySettings extends React.Component {
   constructor (props) {
     super();
 
     this.settings = props.settings;
```
cyyynthia: Using the plugins's settings object in the settings component is deprecated

src/Powercord/plugins/pc-emojiUtility/components/Settings.jsx
```diff
@@ -2,26 +2,61 @@ const { existsSync, lstatSync } = require('fs');
...
-      isCloneIdValid: props.settings.get('defaultCloneId') ? getGuild(props.settings.get('defaultCloneId')) : true,
-      initialCloneIdValue: props.settings.get('defaultCloneId') || null,
+    if (getGuild && getSortedGuilds) {
+      this._setState(false);
+    }
+  }
+
+  _setState (update = true) {
+    const state = {
+      useEmbeds: this.settings.get('useEmbeds', false),
```
cyyynthia: use this.props.getSetting instead

cyyynthia (changes requested):
You also used `this.set` in the settings component, and this is deprecated as well. use `this.props.updateSetting(setting, newValue)` and `this.props.toggleSetting(setting)` instead

cyyynthia:
> 
> 
> shouldn't this plugin be moved into its own repo now?

It'll get moved with most of other internal plugins once the Plugins tab is finished

cyyynthia (approved):

PR #173, opened 2019-04-23
Plugins page improvements

zspri:
Added `Open Plugins Folder` and `Open Themes Folder` links to the plugin settings page, and made the work-in-progress warning look better

### DIFF
--- a/src/Powercord/plugins/pc-pluginManager/scss/style.scss
+++ b/src/Powercord/plugins/pc-pluginManager/scss/style.scss
@@ -7,10 +7,9 @@
     top: 0;
     left: 0;
     right: 0;
-    background-color: #d84040;
     text-align: center;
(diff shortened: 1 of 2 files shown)


src/Powercord/plugins/pc-pluginManager/scss/style.scss
```diff
@@ -7,10 +7,9 @@
     top: 0;
     left: 0;
     right: 0;
-    background-color: #d84040;
     text-align: center;
-    font-size: 18px;
```
cyyynthia: I feel like this change make it less visible than before. Alert was made to make people actually read it (and even rn ppl seem so ignore it lol). Looks nice to me with font-size + padding css properties, maybe consider adding them back

cyyynthia (approved):

PR #181, opened 2019-04-25
fix pc-codeblock errors

MrLar:
Fixes errors when jumping a messages due to the element not having a second child.


cyyynthia (approved):
this is a one change pr brb closing

PR #182, opened 2019-04-25
GitHub Organization Changes

auguwu:
The commit basically changes `powercord-org` to `powercord-community` when installing plugins and/or themes from git's cli component.


cyyynthia:
We won't merge this as it's not how installing plugin and theme will be handled once it'll be a thing.

Also both of those functions are not callable either because it's a comment  or because the UI haven't been done.

On another note, please avoid one change PRs as they just generate spam, unless it's to fix a breaking bug

davidcralph:
no contributor role 4 u hahayeet

PR #184, opened 2019-04-26
Better Cursor for Guilds Plugin (beside Guild itself)

ghostlydilemma:
Very small change and don't know if intended: 

When hovered beside the guild itself and onto the span (or even next to that), cursor grab appears. I don't know if this is intended but it sure looks odd. 
image
vs (fixed with cursor: pointer)
image

### DIFF
--- a/src/Powercord/plugins/pc-guilds/style.scss
+++ b/src/Powercord/plugins/pc-guilds/style.scss
@@ -73,7 +73,7 @@
   }
 }
 
-[data-react-beautiful-dnd-drag-handle="4"]{
+.pc-guild {
   cursor: pointer!important;
 }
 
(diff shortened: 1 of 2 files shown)


src/Powercord/plugins/pc-guilds/style.scss
```diff
@@ -73,6 +73,10 @@
   }
 }
 
+[data-react-beautiful-dnd-drag-handle="4"]{
```
cyyynthia: This attribute may change at any time

cyyynthia (changes requested):

src/Powercord/plugins/pc-guilds/style.scss
ghostlydilemma: Does the suggestions feature work?
I hope so
```suggestion
.pc-guild{
```

ghostlydilemma: Because one space good
```suggestion
.pc-guild {
```

cyyynthia (approved):

PR #188, opened 2019-05-03
added stable & ptb injection

Venipa:


cyyynthia:
Thanks for the PR, but officially decided to not support other release channels than Canary due to the internal differences between releases that are painful to support.

Venipa:
well you actually dont need to support them just make the option available that allows user to inject to their preferred version

regards

MrLar:
That would create the false assumption that they will work without having to modify anything. Which in very many cases is not the case due to the huge internal differences between the release channels.

PR #189, opened 2019-05-03
fix for reloading plugins, added jshintrc to correct es version hints

Venipa:


MrLar:
May I question why you add the config for jshint but not the actual module

Venipa:
my bad tho im using vscode and jshint installed along i guess

MrLar:
It is still a bad practice to assume people already have the module. Besides given internal discussions with bowser im fairly certain that sync dependencies are meant to be handled very differently and not as easy as that given certain circumstances he has not further outlined in our discussions plus future additions for remote dependency fetching would require this to be refactored anyway.

Venipa:
well people creating plugins may want to reload their plugins without reloading the whole window.
true it may or will be refactored but for now i guess its okay for dev use.

MrLar:
Developers can reload their plugins using the dev console of discord

MrLar:
Besides disabling and enabling doesn't even reload the plugin. Only remounting does.

Venipa:
still why not making something work and change it later.
currently its left broken.

cyyynthia:
imagine reading in 2019
who tf does that
that's almost like it's smth smart to do

The only reason the tab is here is because we didn't wanted to disable it to get feedback on the visual design, and because there is no real point on it (and I left it enabled to allow people to see installed plugins, and their registered permissions). We also made an announcement where we mentionned it's broken. Pushing it as-is would break in some scenarios and that's why the feature is "left broken". The PR that changed almost 5k lines of code, so if I wanted to add a fix that is that simple I'd have done it. I just wanted to release it even if that was not working because the PR added a *lot* of features that people wanted and I prefered merge anyway. Keep in mind plugin manager is still a wip that is not even meant to be working.

Venipa:
then i'll close it i guess

PR #191, opened 2019-05-21
Introduce some bug fixes and extra additions

GriefMoDz:
* Replace all traces of the `sendTyping` method with `startTyping` as it is now deprecated
* Set the `top` declaration for the plugin warning found under 'pc-pluginManager' to a value of '-1.4px' - this prevents the ghost pill from cutting off the screen.
* Make the recently implemented plugin/theme folder opener ESLint-compliant (`Installed.jsx`).
* And finally, context menu buttons now react when disabled by shaking the screen (this has been broken for some time now...). I've also made changes to sliders by giving plug-in developers the freedom to customize/use markers and class names.

### DIFF
--- a/src/Powercord/plugins/pc-commands/monkeypatchTyping.js
+++ b/src/Powercord/plugins/pc-commands/monkeypatchTyping.js
@@ -1,14 +1,14 @@
 const { typing } = require('powercord/webpack');
 
 module.exports = async function monkeypatchTyping () {
-  typing.sendTyping = (
-    _sendTyping => (id) =>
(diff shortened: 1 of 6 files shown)


ghostlydilemma:
In case I currently don't mix up everything:
For the: top declaration
I originally intended the -5px to be to remove the top border radius...I think
I cant tell you why I haven't just removed the border radius on top with the normal CSS attribute and kept top set to 0...that would probably be one further addition to your pr

cyyynthia (approved):
This pr have been reviewed by the mobile gang

PR #202, opened 2019-06-10
tell people that stable sucks

katlyn:
Adds notice that we only support canary if Powercord is installed on a different release channel.


ghost:
no u

cyyynthia (approved):
dis accurate

PR #209, opened 2019-06-28
Adds a submenu for the user's top songs.

SacreBleuXCVII:
Requires the `user-top-read` Spotify scope.


cyyynthia (approved):
Proper oauth scopes will be added to the backend soon:tm:

PR #216, opened 2019-07-17
[EmojiUtility] Service is due; fix settings tab

GriefMoDz:


cyyynthia (approved):
I'll consider the fix as good enough:tm: for now, and we'll consider about fully upgrading to new changes (aka folders) soon

PR #223, opened 2019-08-09
Little OCD fix

LandenStephenss:
Whenever you do .help cmd (any command) it says the usage is `/cmd` no matter what. Changed that to use the users prefix

### DIFF
--- a/src/Powercord/plugins/pc-commands/commands/help.js
+++ b/src/Powercord/plugins/pc-commands/commands/help.js
@@ -28,7 +28,7 @@ module.exports = {
     } else {
       const command = powercord.api.commands.commands.find(c => [ c.command, ...c.aliases ].includes(commandName));
       if (!command) {
-        result = `Command \`${commandName}\` not found.`;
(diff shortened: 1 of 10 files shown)


aetheryx:
the prefix should be included in the monospace opening character, and also, there are far more occurances of the prefix being hardwired as `/`, feel free to try to find them all

LandenStephenss:
Well i'm not sure why i would find them all whenever you can just use `powercord.api.commands.prefix` to display the correct prefix. And the line i changed is still displaying the prefix

image

aetheryx:
> there are far more **occurences of the prefix being hardwired as `/`**  

I'm implying there's more bugs exactly like the one you're fixing in this PR, and you might want to fix all of them instead of just this specific occurence 😄

LandenStephenss:
Ah, alright. Would you like me to find them and add them or you got it?

aetheryx:
I'll take a look later tonight when I'm free before I merge, but you can take a look until then

PR #224, opened 2019-08-11
Added disable/enable commands. Also have a plugins

LandenStephenss:
Added disable (disables plugins), enable (does the opposite of disable), and added pluginList (just lists all the plugins installed)


src/Powercord/plugins/pc-commands/commands/disable.js
```diff
@@ -0,0 +1,25 @@
...
+            powercord.pluginManager.disable(args[0])
+        }} else {
+            result = 'You do not have that plugin installed'
+        }
+
+        return { 
+            send: false,
+            result
+        }
+    }
```
GriefMoDz: As already brought up in the '#powercord-core' channel of Powercord's Discord, you should consider putting the autocomplete function, `autocompleteFunc()`, to good use to allow for additional functionality as well as error prevention.

I've gone out of my way and have already made a start on this implementation, however you'll need to remove the condition that checks if the arguments are equal to "pc-commands" as the autocomplete menu now takes care of this.

I'd also kindly suggest that a similar method is put under your "[enable.js](https://github.com/powercord-org/powercord/pull/224/files#diff-264f847bb2bb39fb43385a996a88d46b)" file for the sake of consistency.

```suggestion
    },

    autocompleteFunc (args) {
        const plugins = powercord.pluginManager.getPlugins()
            .sort((a, b) => a < b ? -1 : 1 || 0)
            .map(plugin => powercord.pluginManager.plugins.get(plugin));

        if (args.length > 1) {
            return false;
        }

        return {
            commands: plugins
                .filter(plugin => plugin.pluginID !== 'pc-commands' &&
                    plugin.pluginID.includes(args[0].toLowerCase()))
                .map(plugin => ({
                    command: plugin.pluginID,
                    description: plugin.manifest.description
                }))
                .slice(0, 10),
            header: 'powercord plugin list'
        };
    }
```

src/Powercord/plugins/pc-commands/commands/disable.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+    command: 'disable',
+    description: 'Disable a plugin',
+    usage: '{c} [ Plugin Name ]',
```
GriefMoDz: Although this is very minor, the given argument should be displayed as "plugin ID" (not "plugin name") to avoid confusion.

_This change should be reflected inside of "enable.js" as well!_

```suggestion
    usage: '{c} [pluginID]',
```

src/Powercord/plugins/pc-commands/commands/pluginList.js
```diff
@@ -0,0 +1,19 @@
...
+    description: 'List all the plugins that are installed',
+    usage: '{c}',
+    aliases: ['plugins'],
+    func: (args) => {
+        let result;
+        let plugins = powercord.pluginManager.getPlugins()
+        result = {
+            type: 'rich',
+            title: `List of Plugins (${plugins.length})`,
+            description: `\`${powercord.pluginManager.getPlugins().join('\`\n\`')}\``,
```
GriefMoDz:
```suggestion
            description: `\`${plugins.join('\n')}\``,
```

src/Powercord/plugins/pc-commands/commands/pluginList.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+    command: 'pluginlist',
+    description: 'List all the plugins that are installed',
+    usage: '{c}',
+    aliases: ['plugins'],
+    func: (args) => {
```
GriefMoDz:
```suggestion
    func: () => {
```

src/Powercord/plugins/pc-commands/commands/enable.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+    command: 'enable',
+    description: 'Enable a plugin',
+    usage: '{c} [ Plugin Name ]',
```
GriefMoDz:
```suggestion
    usage: '{c} [pluginID]',
```

GriefMoDz (changes requested):

LandenStephenss:
Got everything fixed! added autocomplete to enable and disable, removed the arguments variable from the function in pluginList.js aswell as actually used `plugins`. If theres anything else let me know! Although you might want to consider readding the check for pc-commands since if someone disables it not knowing what they are doing and doesnt know how to use the devTools then they are stuck without commands

GriefMoDz:
Thanks for that, however two of my suggested changes still remain standing;  the unused arguments variable hasn’t been removed from ‘pluginList.js’ and you haven’t reflected the change of usage for the `enable` command.

As for the conditional check, when disabling “pc-commands” (whether it’d be accidental or on purpose) you’ll notice that its autocomplete functionality doesn’t actually unmount until a soft-restart (Ctrl `+` R)  of the Discord client.

That being said, if for whatever reason the user decides to type “pc-commands” as the arguments for either ‘enable’ or ‘disable’, they can simply re-enable the plug-in again almost as if nothing ever happened.

You might want to also address the following styling rules for ESLint (a linter that Powercord use to correct code structure for better formatting as well as carrying out consistent code practices) that our good friend, Travis, has complained about.

GriefMoDz:
Got bored, so I decided to use this time to change some things around as well as resolve all of the known styling issues for you.

Consider replacing all content of the files that you’ve modified with these ones instead:

> enable.js: https://gist.github.com/GriefMoDz/57bf50999658f8942c88bed60c4e47c9
> disable.js: https://gist.github.com/GriefMoDz/ceb3515ed3c46c46af52366f4f5b6e4a
> pluginList.js (rename to plugin.js): https://gist.github.com/GriefMoDz/0a12d5ffdc9e59d4b853d09aaf74c045

GriefMoDz (approved):

PR #229, opened 2019-09-10
Extended guild classes

### LINKED ISSUE #215: [Feature Request] bring back extended guild selectors
**Describe the feature request**
yes

**What Plugin would it be for?**
yes

**Expected behavior**
yes

cloudrac3r:
Closes #215. This doesn't address the comment about guild names.

If you want to optimise out the ClassList thing into string matching functions then go ahead, but JS is fast so I don't care enough to do it myself.

The render function is called on a guild (and only that guild) when it is hovered or selected, which was very convenient for the creation of this plugin.


cloudrac3r:
fuck you eslint i do what i want

src/Powercord/plugins/pc-guild-classes/index.js
```diff
@@ -1,57 +1,56 @@
...
+
+  async _patchGuild () {
+    const guildInner = await waitFor('.blobContainer-239gwq');
+    const guildElement = guildInner.parentElement;
+    const Guild = getOwnerInstance(guildElement);
+    inject('cadence-guild-classes', Guild.constructor.prototype, 'render', (_, res) => {
+      if (res._owner) {
+        const { hovered } = res._owner.memoizedState;
+        const { selected } = res._owner.pendingProps;
+        const classList = new ClassList(res.props.className);
```
cyyynthia: is an extra class required? Wouldn't something like this enough?
```js
const classNames = [ ...new Set(className.split(' ')) ]
hovered && classNames.push('...')
selected && classNames.push('...')
res.props.className = classNames.join(' ');
```

src/Powercord/plugins/pc-guild-classes/index.js
```diff
@@ -1,57 +1,56 @@
...
+  startPlugin () {
+    this._patchGuild();
+  }
+
+  pluginWillUnload () {
+    uninject('cadence-guild-classes');
+  }
+
+  async _patchGuild () {
+    const guildInner = await waitFor('.blobContainer-239gwq');
```
cyyynthia: Avoid using raw classes in here, fetch them dynamically

src/Powercord/plugins/pc-guild-classes/index.js
```diff
@@ -1,57 +1,56 @@
...
+    const guildElement = guildInner.parentElement;
+    const Guild = getOwnerInstance(guildElement);
+    inject('cadence-guild-classes', Guild.constructor.prototype, 'render', (_, res) => {
```
cyyynthia: Keeping injection names in here with "pc-" would be more consistent as it's an internal plugin

src/Powercord/plugins/pc-guild-classes/index.js
```diff
@@ -1,57 +1,56 @@
...
+      if (res._owner) {
+        const { hovered } = res._owner.memoizedState;
+        const { selected } = res._owner.pendingProps;
+        const classList = new ClassList(res.props.className);
+        classList.conditional(hovered, 'pgc-hovered');
```
cyyynthia: is a prefix really required here? wouldn't just go for .hovered be enough?

cyyynthia (changes requested):

GriefMoDz (approved):

PR #232, opened 2019-09-25
Fix lmgtfy

cloudrac3r:
What does this fix? It fixes not being able to use the command when you only enter a single word. The bug is caused by whatever the hell is going on in line 40.

I also refactored the horror that was line 60.

This could still really do with a full refactor, but I don't care enough to do it myself.


davidcralph:
what are you on about line 60 is the greatest line ever written

cyyynthia (approved):

PR #234, opened 2019-10-03
Various changes

davidcralph:
spoopy month changelog:
1. (env_check.js) "as this will generate conflicts, and bloats your Windows" -> "as this will generate conflicts and bloat your Windows" (sounds better)
2. (pc-clickableEdits) Switched to "module.exports = Class" on index.js to be consistent with Settings.jsx and added semicolon where travis was sad
3. (pc-hastebin, pc-heygirl) Removed full stop to be consistent with the other plugins
4. (Various commands) "[ example ]" -> "[example]" (sounds better, also consistent with other files instead of a mix of the two) 
5. (pc-lmgtfy) Capitalised Google because company names
6. (pc-settings) "This will make Discord slower" -> "This will make temporarily make Discord slower" (informs users that the change isn't perm)
7. (pc-spotify) "Link back Spotify" -> "Relink Spotify" (sounds better and also more consistent with the other messages)
8. (pc-spotify) Added skip alias for next command
9. (pc-tags) Capitalised tags in places where it should be
10. (pc-translate) "lang" -> "language" (sounds better)
11. ~~(reducer.js) "function ()" -> "() =>" (because this is 2019)~~ crab

have fun and inform me if there is anything that needs changing/undoing

~~if ur reading ur cute~~

### DIFF
--- a/src/Powercord/plugins/pc-hastebin/index.js
+++ b/src/Powercord/plugins/pc-hastebin/index.js
@@ -12,7 +12,7 @@ module.exports = class Hastebin extends Plugin {
     this.registerCommand(
       'hastebin',
       [],
-      'Lets you paste content to Hastebin.',
+      'Lets you paste content to Hastebin',
       '{c} [ --send ] < --clipboard | FILE_URL >',
       async (args) => {
         const send = args.includes('--send')
--- a/src/Powercord/plugins/pc-hastebin/index.js
+++ b/src/Powercord/plugins/pc-hastebin/index.js
@@ -13,7 +13,7 @@ module.exports = class Hastebin extends Plugin {
       'hastebin',
       [],
       'Lets you paste content to Hastebin',
-      '{c} [ --send ] < --clipboard | FILE_URL >',
+      '{c} [--send] <--clipboard | FILE_URL>',
       async (args) => {
         const send = args.includes('--send')
           ? !!args.splice(args.indexOf('--send'), 1)
--- a/src/Powercord/plugins/pc-settings/components/GeneralSettings.jsx
+++ b/src/Powercord/plugins/pc-settings/components/GeneralSettings.jsx
@@ -118,7 +118,7 @@ module.exports = class GeneralSettings extends React.Component {
         </Category>
 
         <ButtonItem
-          note={'Removes everything stored in Discord\'s cache folder. This will make temporarily make Discord slower, as all resources will have to be fetched again.'}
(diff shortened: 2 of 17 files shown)


src/Powercord/plugins/pc-hastebin/index.js
```diff
@@ -12,7 +12,7 @@ module.exports = class Hastebin extends Plugin {
     this.registerCommand(
       'hastebin',
       [],
-      'Lets you paste content to Hastebin.',
+      'Lets you paste content to Hastebin',
       '{c} [ --send ] < --clipboard | FILE_URL >',
```
GriefMoDz: Might as well close up the spacing between these arguments for additional sake of consistency.
```suggestion
      '{c} [--send] <--clipboard | FILE_URL>',
```

src/Powercord/plugins/pc-settings/components/GeneralSettings.jsx
```diff
@@ -118,7 +118,7 @@ module.exports = class GeneralSettings extends React.Component {
         </Category>
 
         <ButtonItem
-          note={'Removes everything stored in Discord\'s cache folder. This will make Discord slower, as all resources will have to be fetched again.'}
+          note={'Removes everything stored in Discord\'s cache folder. This will make temporarily make Discord slower, as all resources will have to be fetched again.'}
```
GriefMoDz: Looks like you've put one too many "make"'s here. Let's fix that! 😊
```suggestion
          note={'Removes everything stored in Discord\'s cache folder. This will temporarily make Discord slower, as all resources will have to be fetched again.'}
```

src/Powercord/plugins/pc-spotify/commands/play.js
```diff
@@ -3,7 +3,7 @@ const urlRegex = /\/track\/([A-z0-9]*)/;
 module.exports = {
   command: 'play',
   description: 'Play a Spotify URL',
-  usage: '{c} <URL>',
+  usage: '{c} < URL >',
```
GriefMoDz: Here we can do the same and close up the spacing before and after "URL".
```suggestion
  usage: '{c} <URL>',
```

GriefMoDz (changes requested):

davidcralph:
Thanks for this @GriefMoDz

GriefMoDz (approved):

davidcralph:
:tada:

PR #237, opened 2019-10-13
UI coloring redo using CSS Variables

ghostlydilemma:
As Discord implemented CSS Variables for most of the text and backgrounds, using those in Powercord would be a good idea. (Both because of current themes utilizing those and (maybe?) future Discord in house themes).

### DIFF
--- a/src/Powercord/managers/styles/css/_settings.scss
+++ b/src/Powercord/managers/styles/css/_settings.scss
@@ -98,12 +98,5 @@ div[class*='switchItem-'] label button {
   text-transform: uppercase;
   letter-spacing: .5px;
   height: 16px;
-}
-
-.theme-dark .powercord-title {
-  color: #b9bbbe;
-}
-
-.theme-light .powercord-title {
-  color: #72767d;
+  color: var(--channels-default);
 }
--- a/src/Powercord/plugins/pc-codeblocks/style.scss
+++ b/src/Powercord/plugins/pc-codeblocks/style.scss
@@ -24,14 +24,8 @@
   }
 }
 
-.theme-light {
-  .powercord-codeblock-lang {
-    color: #777777a8 !important;
-  }
-}
-
 .powercord-codeblock-lang {
-  color: #ffffff;
+  color: var(--text-muted);
   border-bottom: 1px solid rgba(0, 0, 0, 0.2);
 
   line-height: 20px;
--- a/src/Powercord/plugins/pc-codeblocks/style.scss
+++ b/src/Powercord/plugins/pc-codeblocks/style.scss
@@ -25,7 +25,7 @@
 }
 
 .powercord-codeblock-lang {
-  color: var(--text-muted);
+  color: var(--text-normal);
   border-bottom: 1px solid rgba(0, 0, 0, 0.2);
 
   line-height: 20px;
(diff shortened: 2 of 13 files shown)


src/Powercord/managers/styles/css/_settings.scss
```diff
@@ -98,12 +98,5 @@ div[class*='switchItem-'] label button {
...
-}
-
-.theme-dark .powercord-title {
-  color: #b9bbbe;
-}
-
-.theme-light .powercord-title {
-  color: #72767d;
-}
+  color: var(--channels-default);
```
GriefMoDz, line 101: I don't know about you, but I'd probably opt-in for using `--header-secondary` here instead of `--channels-default` as it's a little more easy on the :eyes: and is somewhat fitting for its purpose.

```suggestion
  color: var(--header-secondary);
```

src/Powercord/plugins/pc-codeblocks/style.scss
```diff
@@ -24,14 +24,8 @@
...
 
-.theme-light {
-  .powercord-codeblock-lang {
-    color: #777777a8 !important;
-  }
-}
-
 .powercord-codeblock-lang {
-  color: #ffffff;
+  color: var(--text-muted);
```
GriefMoDz: Uh, same case here. I'd personally use `--text-normal` which is very close to the traditional white for dark theme and a far more standing out black for light themed users.

```suggestion
  color: var(--text-normal);
```

src/Powercord/plugins/pc-pluginManager/scss/_plugin.scss
```diff
@@ -77,30 +79,18 @@
     }
   }
 
+  &-permissions, &-footer {
+    border-top: 1px rgba(0,0,0,0.2) solid;
```
GriefMoDz: Maybe try using `--background-modifier-accent` here to match the color of Discord's settings dividers... :man_shrugging: 

```suggestion
    border-top: 1px var(--background-modifier-accent) solid;
```

src/Powercord/plugins/pc-updater/style.scss
```diff
@@ -55,7 +55,7 @@
 
     .button {
```
GriefMoDz, line 56: It'd be great to introduce a new rule-set here for the `:hover` pseudo-class so that you can distinguish which of the four buttons from the update toast you're hovering over. I suggest something like this would do just fine:
```css
.button {
  ...

  &:hover {
    background-color: var(--background-modifier-hover);
  }
}
```

GriefMoDz (changes requested):

src/Powercord/managers/styles/css/_settings.scss
ghostlydilemma: I tried to get as close to the default Discord look as possible: the h5's in the discord settings (oddly enough) use `--channels-default` for the coloring. I thought about `--header-secondary`, but it seems a little bit too dark, imo.
image
image

src/Powercord/plugins/pc-pluginManager/scss/_plugin.scss
ghostlydilemma: True that looks much better, only later in the code found the background-modifiers myself, that was sadly before that point. Thanks for the suggestion :)

src/Powercord/plugins/pc-updater/style.scss
ghostlydilemma: Great suggestion, will add that asap

src/Powercord/managers/styles/css/_settings.scss
GriefMoDz: Very well. I'll mark this as resolved. Thanks for your input.

GriefMoDz (approved):

PR #240, opened 2019-10-14
React Dev Tools without needing it installed in Chrome

rosethornbush:
Implementation of a standalone react-dev-tools installation in powercord.

### DIFF
--- a/src/Powercord/plugins/pc-reactdevtools/index.js
+++ b/src/Powercord/plugins/pc-reactdevtools/index.js
@@ -1,6 +1,6 @@
 const { join } = require('path');
 const { remote } = require('electron');
-const { createWriteStream, existsSync } = require('fs');
+const { createWriteStream, existsSync, readFileSync } = require('fs');
 const { Plugin } = require('powercord/entities');
 const { get } = require('powercord/http');
 const unzip = require('unzip-crx');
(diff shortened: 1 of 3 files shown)


src/Powercord/plugins/pc-reactdevtools/index.js
```diff
@@ -1,60 +1,76 @@
-const { resolve, join } = require('path');
-const { remote } = require('electron');
-const { readdir } = require('fs').promises;
-const { Plugin } = require('powercord/entities');
+const { join } = require("path");
+const { remote } = require("electron");
+const { createWriteStream, existsSync } = require("fs");
+const { Plugin } = require("powercord/entities");
+const request = require("request");
```
cyyynthia: Use the internal powercord/http package instead

src/Powercord/plugins/pc-reactdevtools/index.js
```diff
@@ -1,60 +1,76 @@
...
+      .getCurrentWindow()
+      .webContents.removeListener("devtools-opened", this.listener);
   }
 
-  async listener () {
-    remote.BrowserWindow.removeDevToolsExtension('React Developer Tools');
-
-    let version;
-    let ok = true;
+  async listener() {
```
cyyynthia: async here is useless as no async operations are performed in here anymore.

src/Powercord/plugins/pc-reactdevtools/index.js
```diff
@@ -1,60 +1,76 @@
...
+          );
+        });
+      }
+    );
+  }
+
+  startPlugin() {
     this.listener = this.listener.bind(this);
-    remote.getCurrentWindow().webContents.on('devtools-opened', this.listener);
+    this.download();
```
cyyynthia: This will call download on every startup, and while it's a good solution for keeping it up to date i don't feel that should be left as-is. Maybe make it conditional if not installed or something?

src/Powercord/plugins/pc-reactdevtools/index.js
```diff
@@ -1,60 +1,76 @@
...
+        file.write(body, err => {
+          if (err) {
+            this.error(err);
+          }
+          file.close();
+        });
+        unzip(this.path, this.folderPath).then(() => {
+          this.listener();
+          this.log(
+            "If you are unable to find the tabs for React Developer Tools in the Developer Tools, Reload your client (ctrl + r)"
```
cyyynthia: This will get called everytime the plugin is loaded as well (see below). Also reloading shouldn't be technically required as the extension gets reloaded everytime devtools are opened.

cyyynthia (changes requested):

cyyynthia:
Also didn't mentionned it but as you can see in the Travis build the linter isn't happy rn

src/Powercord/plugins/pc-reactdevtools/index.js
rosethornbush: I have noticed on first install of react dev tools whether by my method or the old method of using chrome, me and friends have had to reload to have the tabs appear.

src/Powercord/plugins/pc-reactdevtools/index.js
rosethornbush: I didn't even see that, thanks for letting me know. It seems to be part of the original code for the plugin.

src/Powercord/plugins/pc-reactdevtools/index.js
rosethornbush: Furthermore, the log only fires when the plugin downloads/updates which would only happen on first time install or there's an update to the plugin so it really isn't too intrusive.

rosethornbush: Seeing that I have resolved calling download on each plugin start which causes this log to happen on every start, I am resolving this.

GriefMoDz (approved):
Looks good to me! - here's my seal of approval.

Now just waiting on Bowser to review these changes/run over some his own extra checks and then we should be on our way for merge. :+1:

cyyynthia (approved):

PR #244, opened 2019-10-24
A few fixes for the hastebin plugin

ghostlydilemma:
Changes the default from Aeths dead domain to the official hastebin.
And since we all know how stable it is: I implemented a small check whether the post was sucessful, if not it displays a small local warning.


aetheryx:
use https://hasteb.in/

cyyynthia (approved):

PR #246, opened 2019-10-25
Tweaks to the Toast and Updater

tale:
1. Changed some CSS with the Updater and added an entry to show the install path of Powercord
2. Added an ID system to the toast buttons so that we can use custom colors or styles
3. Changed the update toast to use the blue ID in the 'Update now' toast
4. Changed the toast to better fit with Discord styling

If you have any questions contact me on Discord: Flexstar#0834

### DIFF
--- a/src/Powercord/managers/styles/css/_updater.scss
+++ b/src/Powercord/managers/styles/css/_updater.scss
@@ -70,6 +70,40 @@
     }
   }
 
+  #blue {
+
+    background-color: #7289DA;
+    color: #ffffff;
+
+    &:hover {
+
+      background-color: #677BC4;
+    }
+  }
+
+  #red {
+
+    background-color: rgba(0,0,0,0);
+    border: 1px solid rgba(240,71,71,.3);
+    color: #F04747;
+
+    &:hover {
+
+      border-color: rgba(240,71,71,.6);
+    }
+  }
+
+  #yellow {
+
+    background-color: #FAA61A;
+    color: #ffffff;
+
+    &:hover {
+
+      background-color: #E19157;
+    }
+  }
+
   &.leaving {
     animation: slideout 0.7s forwards;
   }
--- a/src/Powercord/managers/styles/css/_updater.scss
+++ b/src/Powercord/managers/styles/css/_updater.scss
@@ -22,11 +22,12 @@
   display: flex;
   flex-direction: column;
   justify-content: center;
-  background-color: var(--background-tertiary);
+  background-color: var(--background-secondary);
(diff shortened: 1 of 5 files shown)


src/Powercord/managers/styles/css/_updater.scss
```diff
@@ -69,14 +70,48 @@
     }
   }
 
+  #blue {
```
cyyynthia, line 73: IDs for styling is an extremely bad practice and can lead in breaking standards as, according to the HTML standard, there must be only 1 and only 1 element with a given id. Using such non-specific IDs can lead to breaking that standard quickly. Also, theming on IDs is something I personally don't like. Though before rewriting this, check the comment below concerning the use of custom made elements.

src/Powercord/plugins/pc-updater/components/Settings.jsx
```diff
@@ -72,17 +73,21 @@ module.exports = class UpdaterSettings extends React.Component {
...
-            <span>Revision:</span>
+            <span className="span-header">Revision:</span>
             <span>{powercord.gitInfos.revision.substring(0, 7)}</span>
           </div>
           <div>
-            <span>Branch:</span>
+            <span className="span-header">Branch:</span>
             <span>{powercord.gitInfos.branch}</span>
           </div>
+          <div>
```
cyyynthia, line 87: tbh I'm unsure if I like it or not, showing a path in such a small space looks weird to me. Might be personal preferences, anyone else to share their thaughts?

src/fake_node_modules/powercord/components/Toast.jsx
```diff
@@ -59,7 +59,7 @@ module.exports = class Toast extends React.Component {
 
           <div className='buttons'>
             {state.buttons.filter(b => b).map(button => (
-              <div className='button' onClick={() => button.onClick(this.remove.bind(this))}>
+              <div className='button' id={button.type} onClick={() => button.onClick(this.remove.bind(this))}>
```
cyyynthia, line 62: ew IDs

Also at this pont, why not just using Discord's components instead of copying CSS, for both the whole thing and buttons? You can just use Card and Button to get the same style as Discord's without effort, and this will let themes immediately work on custom Powercord elements (as they're basically Discord elements).

You can then let buttons define "color" and "look", and just throw it in the button component and let Discord apply its own styles. You can then add very little amount of css to finalize the button element, like its width or height.

cyyynthia (changes requested):

tale:
I'll be closing this as @GriefMoDz or someone else seems to be working on toasts

PR #247, opened 2019-10-25
Make the updater not take 10 hours

cloudrac3r:
- Settings API concurrent write protection is no longer bowserware
- Extract magic fields to updater concurrency number, currently set to Infinity

There were no issues with updating all plugin repos at the same time with my legacy plugin updater extension, so why are you checking for updates in small batches here?

Eslint is screaming at me, but I don't know how to fix that.

Works on my machine.

### DIFF
--- a/src/Powercord/apis/settings/store/index.js
+++ b/src/Powercord/apis/settings/store/index.js
@@ -63,18 +63,34 @@ class SettingsStore extends Flux.Store {
     return Object.keys(this.getSettings(category));
   }
 
+  static pending = [];
+  static emptying = false;
+
   static async _persist (category, settings) {
     // Let's not write concurrently
-    while (this._write_locked) {
-      await sleep(10);
+    SettingsStore.pending.push({ category, settings });
+    SettingsStore._triggerQueue();
+  }
+
+  static _triggerQueue () {
+    if (!SettingsStore.emptying) {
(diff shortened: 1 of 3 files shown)


cyyynthia:
I did set concurrency limit to prevent firing up a lot of process and causing noticable performance drops. As the updater frequently runs (Every 15 minutes by default), it is a concern to make it not slow everything down. My CPU fans started being noticably louder when running all plugin updates concurrently, which should *not* happen for a frequent background task. And I have an i7 so imagine for weak CPUs

davidcralph:
~~can we ask visual studio code to get a concurrency limit so it doesn't open 250 copies of git all the fucking time~~

cyyynthia (changes requested):
other than concurrency lgtm (yes bowserware at 2am without enough caffeine)

src/Powercord/apis/settings/store/index.js
```diff
@@ -63,18 +63,34 @@ class SettingsStore extends Flux.Store {
     return Object.keys(this.getSettings(category));
   }
 
+  static pending = [];
```
cyyynthia: "but I don't know how to fix that."
how about not using Node 12+ features on a Node 10 project

cyyynthia (approved):

PR #254, opened 2019-10-29
Custom Routes API

cyyynthia:
Let plugins register new routes, letting them build more complex UIs 🎉 

## New API
in plugins: `this.registerRoute('/path/you/want', ReactComponent)`
 => Registers the route `/_powercord/path/you/want` (don't forget the `_powercord` when linking!) with the component ReactComponent inside

## What is left to do
 - Let plugins decide if they want to get the sidebar (channels + account) or not
 - Error boundaries for dumb people like me

## Small changes that also got into this
 - iframe protections headers are now yeeted
 - a new "loaded" event is emitted by powercord (it was supposed to be for the router api but i don't use it and why not keep the event anyway)

If no one objects i'll merge this tomorrow 👀


NurMarvin:
I approve this

PR #258, opened 2019-11-08
Little adjustments to findInTree

intrnl:


cyyynthia (approved):
This was reviewed by the phone in class gang

PR #264, opened 2019-11-18
Cache JSX transformations

intrnl:
Improves subsequent JSX requires by [up to 48 percent (personal test, may vary)](https://haste.powercord.dev/muhapohove.txt)

### DIFF
--- a/src/Powercord/modules/jsx.js
+++ b/src/Powercord/modules/jsx.js
@@ -17,14 +17,44 @@
  */
 
 const sucrase = require('sucrase');
-const { readFileSync } = require('fs');
+const { join } = require('path');
+const { readFileSync, promises: { mkdir, writeFile } } = require('fs');
+const { createHash } = require('crypto');
 
-module.exports = () =>
-  require.extensions['.jsx'] = (_module, filename) =>
-    _module._compile(
-      sucrase.transform(readFileSync(filename, 'utf8'), {
+const cacheDir = join(__dirname, '../../../cache/jsx/');
+
+const checksum = (str) => createHash('sha1').update(str).digest('hex');
+
+module.exports = () => {
+  mkdir(cacheDir, { recursive: true });
+
+  require.extensions['.jsx'] = (_module, filename) => {
+    const source = readFileSync(filename, 'utf8');
+    const hash = checksum(`/* jsx-sucrase | ${filename} */${source}`);
+    const transformPath = join(cacheDir, `${hash}.js`);
+
+    let alreadyTransformed = false;
+    let transform;
+    try {
+      transform = readFileSync(transformPath, 'utf8');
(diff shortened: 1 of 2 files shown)


intrnl:
This caching implementation does not have a mapping of file names to their respective cache files, so this will be painful when developing as suddenly there would be a lot of trash generated.

Since I don't think adding mappings would add any performance penalties, I'm thinking of baking that in right now. 👍

intrnl:
Welp, I don't think having mappings is worth the effort considering the edge cases that comes along with it. So I'd say this is good to go.

intrnl:
Well it seems that the CI is failing thanks to an unrelated file

Zerthox:
The additional generated trash during development should be pretty small, since the only scenarios where it'll transform the same content is when the file has been modified but the contents haven't changed or have been reverted to an already cached state.

For further trash reduction the following ideas might be worth considering:
1. Make the cache clean itself. Remove already cached instances of a file on transform. This would also remove old files when Powercord updates, but it requires mappings.
2. Add an option to disable caching. This might be worth implementing for users who are short on disk space aswell.
3. Add a command to clear (or clean) the cache. This would be a little annoying to use during development, but it would also give end users more control over the cache.

EDIT: @intrnl you forgot to revert the cache name in the gitignore aswell

src/Powercord/modules/jsx.js
```diff
@@ -17,14 +17,40 @@
...
-const { readFileSync } = require('fs');
+const { join } = require('path');
+const { readFileSync, existsSync, mkdirSync, writeFile } = require('fs');
+const { createHash } = require('crypto');
 
-module.exports = () =>
-  require.extensions['.jsx'] = (_module, filename) =>
-    _module._compile(
-      sucrase.transform(readFileSync(filename, 'utf8'), {
+const cacheDir = join(__dirname, '../../../cache/jsx/');
```
cyyynthia: gitignore mentions `.cache` but you use `cache`
```suggestion
const cacheDir = join(__dirname, '../../../.cache/jsx/');
```

src/Powercord/modules/jsx.js
```diff
@@ -17,14 +17,40 @@
...
-      }).code,
-      filename
-    );
+      }).code;
+
+      _module._compile(res, filename);
+
+      writeFile(cached, res, (err) => {
+        if (err) {
+          console.error('[JSX]', 'Failed to write to cache');
```
cyyynthia: Pretty:tm: logging
```suggestion
          console.error('%c[Powercord:JSX]', 'color: #7289da', 'Failed to write to cache');
```

cyyynthia (changes requested):

cyyynthia (approved):

PR #271, opened 2019-12-29
imagine using normalized classes

NurMarvin:
smh @Bowser65


cyyynthia:
I'm not the one who wrote this and Harley did most of the job about making Powercord not use .pc- classes

For once that it's not my shitcode don't smh me reee

PR #292, opened 2020-02-02
pt-PT things are gone

resucutie:
all the pt-PT things are gone

if I forget something tell me


cyyynthia:
As I said previously, some strings already have been updated through Weblate and therefore this is just creating conflicts. Due to that doing pull requests to update strings is discouraged as it's harder for us to maintain.

Also, in the future please target v2-dev instead of stable v2.

resucutie:
ok

PR #295, opened 2020-02-12
Remove junk from patcher

intrnl:
have only been tested on my machine™

requires testing from others


intrnl:
Regarding commit `1408e4e`,

```js
class A {}
A.one = 1

class B extends A {}
B.one
// -> 1

A.two = 2
B.two
// -> 2
```

intrnl:
Okay, so this seems to break the ability to mess around with the updater, found an alternate solution though, will commit once I have enough rest.

cyyynthia (approved):

PR #297, opened 2020-02-19
More visible GitHub icon.

SacreBleuXCVII:
As Olykir also mentioned, the icon in the profile modal isn't very visible so the white icon would be better on the black background.


GriefMoDz:
I’ll be closing this as an alternative fix (thanks @theGordHoard for the idea!) has landed under https://github.com/powercord-org/powercord/commit/706d8e2ccbffaa57329955ec64aa9c1dc7449aff.

PR #298, opened 2020-02-23
Auto-inject based on current Discord instances

110Percent:
Injection on Linux does not currently support Discord instances running from locations other than those specified in the injection script. This aims to fix that by looking for the location of the currently-running instance of DiscordCanary using the `ps` command and using it as a base for injection.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
   },
   "homepage": "https://github.com/powercord-org/powercord#readme",
   "dependencies": {
-    "bluebird": "^3.7.2",
     "less": "^3.10.3",
     "node-watch": "^0.6.3",
     "ps-node": "^0.1.6",
--- a/package.json
+++ b/package.json
@@ -22,8 +22,10 @@
   },
   "homepage": "https://github.com/powercord-org/powercord#readme",
   "dependencies": {
+    "bluebird": "^3.7.2",
     "less": "^3.10.3",
     "node-watch": "^0.6.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -22,8 +22,10 @@
   },
   "homepage": "https://github.com/powercord-org/powercord#readme",
   "dependencies": {
+    "bluebird": "^3.7.2",
```
davidcralph: -1 for using a gigantic module

110Percent:
My bad, force of habit. Updated and replaced Bluebird's promisify with the built-in util method.

davidcralph (approved):

cyyynthia:
Or you can even drop ps-node dependency by using this
```js
const { execSync } = require('child_process')
const process = execSync('ps x').toString()
  .split('\n').map(s => s.split(' ').filter(Boolean))
  .find(p => p[4] && p[4].endsWith('DiscordCanary') && p.includes('--type=renderer'))

if (!process) {
  // Discord isn't launched
  return
}
const discordPath = process[4]
```

110Percent:
Tested and works on my end. Glad to see this could be done without a need for any external dependencies at all! :)

aetheryx:
shadowing process is a bad idea, i'm not quite sure why eslint is okay with that

110Percent:
Good call. How's this?

aetheryx:
lgtm

PR #302, opened 2020-03-07
Fix transparency(Linux)

ghost:
Leave titlebar enabled on Linux


MasicoreLord:
That'd also make a frame appear for macOs users, which isn't what you'd want.

ghost:
I'll make sure this only applies to Linux.

PR #312, opened 2020-03-28
OCD Fix but on the right branch

LandenStephenss:


src/Powercord/plugins/pc-tags/commands/list.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
       send: false,
       result: {
         type: 'rich',
-        title: `You have ${keys.length} tags available:`,
+        title: `You have ${keys.length} {keys.length > 1 ? "tags" : "tag"} available:`,
```
cyyynthia: Use single quotes instead of double quotes

cyyynthia (changes requested):

cyyynthia (approved):

PR #321, opened 2020-05-02
Add isInjected to the injector

MeguminSama:
Allows plugins to check if something is injected, so that they can avoid spamming the console with "injection id is already used" if they need to reinject but aren't sure if already injected beforehand for whatever reason.

needed since `injections: []` won't update when imported into other modules.


MeguminSama:
bump. implementing your own state checker per plugin is a pain :^)

PR #325, opened 2020-05-20
Fix glasscord installation

StavZ:
When I tried to set glasscord to powercord by running `npm run glasscord`, 
I got an error `Unidentified argument undefined!`
This a fix for this error :>


cyyynthia:
Glasscord install shouldn't be done using this script anyway (i forgot to delete it lol), please refer to Glasscord's repo for the proper installation steps to follow

cyyynthia:
Also the lack of this argument was intentional and should be specified when executing the command

PR #338, opened 2020-06-05
WIP: Get rid of Electron's remote module

### LINKED ISSUE #337: Get rid of Electron's remote
Just now, Discord released an update **forcefully disabling** Electron's remote module. This was fixed a few minutes later in Powercord through f8180d3ca86f027f4d3102b3b34c3f774239d1a9, however this raises a severe concern over this module we should take into consideration as soon as possible.

For the next couple days and probably months, we're pretty safe leaving this enabled, however Electron announced they'll completely get rid of it in next releases, which will have the same impact but won't be able to be hotfixed in a few mins.

While getting rid of the remote in core Powercord modules should be relatively easy and straightforward, getting rid of it on other places like plugins might be another story. For example, some plugins makes use of the remote to create BrowserWindow (like Multitask or Powercord's SDK).

Right now let's consider we are in the investigating part, where we have to evaluate the amount of work required to get rid of remote, what Discord APIs provide (might be interesting to give a look at the `DiscordNative` global), and what APIs we might need to implement for other plugins.

TTtie:
This PR removes all the code related to the `remote` function of Electron, as described in #337 
Closes #337 
**To do**:
- [ ] Main process injection
- [ ] A simple to use API to create browser windows
- [ ] Split the IPC code into a separate module

### DIFF
--- a/src/patcher.js
+++ b/src/patcher.js
@@ -127,6 +127,18 @@ app.once('ready', () => {
 
 // TODO(TTtie): Move this into another module
 ipcMain.on('pc-getPreload', (ev) => ev.returnValue = originalPreload);
+ipcMain.on('pc-getWebPreferences', (ev) => ev.returnValue = ev.sender.getWebPreferences())
+ipcMain.on('pc-getMaximized', (ev) => {
+  const win = BrowserWindow.fromWebContents(ev.sender);
+  if (!win) return;
+  ev.returnValue = win.isMaximized();
+});
+ipcMain.on('pc-handleMaximize', (ev) => {
+  const win = BrowserWindow.fromWebContents(ev.sender);
+  if (!win) return;
+  win.on('maximize', () => ev.reply('pc-windowMaximize'));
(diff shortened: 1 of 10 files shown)


src/patcher.js
```diff
@@ -69,7 +71,8 @@ class PatchedBrowserWindow extends BrowserWindow {
     }
 
     opts.webPreferences.enableRemoteModule = true;
-    return new BrowserWindow(opts);
+    const win = new BrowserWindow(opts);
```
cyyynthia: Any specific reason for assigning a variable and returning it instead of straight up returning the BrowserWindow?

TTtie: Actually none, I forgot to revert that while I was figuring out how to get the original preload file loaded.

cyyynthia (commented):
I am not a big fan of having an API that specific, other plugins might make use of BrowserWindows and it might be worth making a more generic API other plugins can use instead of tying everything to the splash screen specifically

TTtie:
I'll add that later, that looks like a good idea; respectively, I will add it once I split the IPC into a separate module

PR #340, opened 2020-06-11
Update pc-reactdevtools

Kyza:
It's still a temporary fix considering [this](https://github.com/KyzaGitHub/powercord/blob/2311b3fb83522d11e266a818701bbba15493734b/src/Powercord/plugins/pc-reactdevtools/index.js#L17), but it fixes a several problems with the plugin.

- [x] Actually waits for the CRX to finish downloading before unzipping it.
  - *sigh*
- [x] Doesn't re-add the devtools extension every single time the devtools are opened.
  - That caused problems and made it harder to get the React Devtools tabs to even show up in the first place.
- [x] Uses a promise for the check on whether adding the extension failed or not instead of an `if` statement.
- [x] Purges the React Devtools folder before unzipping the CRX.
  - I'm not *exactly* sure how [`unzip-crx`](https://github.com/peerigon/unzip-crx) works, but this could prevent some--maybe unlikely but still possible--future problems.
- [x] ~~Cleans up the temporary CRX file when it's finished unzipping it.~~

I wasn't sure about the reason for [`path`](https://github.com/powercord-org/powercord/blob/f8180d3ca86f027f4d3102b3b34c3f774239d1a9/src/Powercord/plugins/pc-reactdevtools/index.js#L11) and [`folderPath`](https://github.com/powercord-org/powercord/blob/f8180d3ca86f027f4d3102b3b34c3f774239d1a9/src/Powercord/plugins/pc-reactdevtools/index.js#L15) not being constant variables, so I moved them outside of the plugin.

### DIFF
--- a/src/Powercord/plugins/pc-reactdevtools/index.js
+++ b/src/Powercord/plugins/pc-reactdevtools/index.js
@@ -1,104 +1,104 @@
-const { join } = require("path");
-const { remote } = require("electron");
+const { join } = require('path');
+const { remote } = require('electron');
 const {
-	createWriteStream,
-	existsSync,
-	readFileSync,
-	unlinkSync,
-} = require("fs");
-const { Plugin } = require("powercord/entities");
-const { get } = require("powercord/http");
+  createWriteStream,
+  existsSync,
+  readFileSync,
+  unlinkSync
+} = require('fs');
+const { Plugin } = require('powercord/entities');
+const { get } = require('powercord/http');
 
-const unzip = require("unzip-crx");
+const unzip = require('unzip-crx');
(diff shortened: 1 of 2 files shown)


Kyza:
It redownloads the CRX every time since I forgot that's how it bases the checks. Gimmie a minute.

Thanks, A5rocks.

cyyynthia (changes requested):
Just needs to comply with out linting rules now

Kyza:
~~File size increased. :eyes:~~

The linting errors are fixed.

cyyynthia:
Suppressed by other changes

PR #350, opened 2020-06-28
Like command

StavZ:
I added a like command to pc-spotify plugin which allows add current playing song to liked songs

### DIFF
--- a/src/Powercord/plugins/pc-spotify/commands/like.js
+++ b/src/Powercord/plugins/pc-spotify/commands/like.js
@@ -5,7 +5,7 @@ module.exports = {
   command: 'like',
   description: 'Like the current playing song',
   async executor () {
-    if (!powercord.account && !powercord.account.spotify) {
+    if (!powercord.account || !powercord.account.spotify) {
       return {
         send: false,
         result: 'You need a Powercord account and connected Spotify account to use this feature!'
@@ -22,7 +22,7 @@ module.exports = {
     SpotifyAPI[body[0] ? 'removeSong' : 'addSong'](currentTrack.id);
     return {
       send: false,
-      result: `You ${body[0] ? 'removed' : 'added'} **${currentTrack.name}** by **${currentTrack.artists}** from your Liked Songs.`
+      result: `You ${body[0] ? 'removed' : 'added'} **${currentTrack.name}** by **${currentTrack.artists}** ${body[0] ? 'from' : 'to'} your Liked Songs.`
     };
   }
 };
--- a/src/Powercord/plugins/pc-spotify/commands/like.js
+++ b/src/Powercord/plugins/pc-spotify/commands/like.js
@@ -1,0 +1,29 @@
+const playerStore = require('../playerStore/store');
+const SpotifyAPI = require('../SpotifyAPI');
+
+module.exports = {
+  command: 'like',
+  description: 'Like a current playing song',
+  async executor () {
+    const currentTrack = playerStore.getCurrentTrack();
+    if (!currentTrack) {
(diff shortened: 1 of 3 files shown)


cyyynthia (changes requested):
The command doesn't take into account if the account is from Discord or Powercord. We can only update someone's library if they linked their account to Powercord since we need some extra scopes Discord doesn't request. See [this](https://github.com/powercord-org/powercord/blob/v2-dev/src/Powercord/plugins/pc-spotify/components/Modal.jsx#L160-L168) for an example on how to check this.

StavZ:
Ok. I'll fix it

src/Powercord/plugins/pc-spotify/commands/like.js
```diff
@@ -12,6 +12,14 @@ module.exports = {
         result: 'You are not currently listening to anything.'
       };
     }
+    const hasCoolFeatures = powercord.account && powercord.account.spotify;
+    if (!hasCoolFeatures) {
+      return {
+        send: false,
+        result: 'You need a [powercord](https://powercord.dev/) account and connected Spotify account.'
```
davidcralph: Powercord should be capitalised here 

(i just realised i commented on an outdated version of the file, but this issue still applies)

davidcralph (approved):

src/Powercord/plugins/pc-spotify/commands/like.js
```diff
@@ -0,0 +1,37 @@
...
+        result: 'You are not currently listening to anything.'
+      };
+    }
+    const hasCoolFeatures = powercord.account && powercord.account.spotify;
+    if (!hasCoolFeatures) {
+      return {
+        send: false,
+        result: 'You need a Powercord account and connected Spotify account.'
+      };
+    }
```
cyyynthia: It'd be better to check for the Powercord account first
```suggestion
    if (!powercord.account && powercord.account.spotify) {
      return {
        send: false,
        result: 'You need a Powercord account and connected Spotify account.'
      };
    }
    const currentTrack = playerStore.getCurrentTrack();
    if (!currentTrack) {
      return {
        send: false,
        result: 'You are not currently listening to anything.'
      };
    }
```

src/Powercord/plugins/pc-spotify/commands/like.js
```diff
@@ -0,0 +1,37 @@
...
+
+    const { body } = await SpotifyAPI.checkLibrary(currentTrack.id);
+    if (body[0]) {
+      SpotifyAPI.removeSong(currentTrack.id);
+      return {
+        send: false,
+        result: `You removed **${currentTrack.name} by ${currentTrack.artists}** from your Liked Songs`
```
cyyynthia: Styling nitpicking
```suggestion
        result: `You removed **${currentTrack.name}** by **${currentTrack.artists}** from your Liked Songs`
```

src/Powercord/plugins/pc-spotify/commands/like.js
```diff
@@ -0,0 +1,37 @@
...
+      };
+    }
+    SpotifyAPI.addSong(currentTrack.id);
+    return {
+      send: false,
+      result: `You added **${currentTrack.name} by ${currentTrack.artists}** to your Liked Songs`
```
cyyynthia: Styling nitpicking
```suggestion
      result: `You added **${currentTrack.name}** by **${currentTrack.artists}** to your Liked Songs`
```

cyyynthia (changes requested):

cyyynthia (approved):

PR #354, opened 2020-07-14
Sudo warning for Linux

Weilbyte:
When running the injector without root privileges under Linux, it will throw an `EACCES` error. 

This simple change tells the users to run as sudo instead of throwing the error at them.


intrnl (changes requested):
Make it generic for all platforms and as a suggestion.

Weilbyte:
Closing this one as I have selected the wrong branch (oops).

PR #355, opened 2020-07-14
Sudo warning for Linux

Weilbyte:
When running the injector without root privileges under Linux, it will throw an EACCES error.

This simple change tells the users to run as sudo instead of throwing the error at them.

### DIFF
--- a/injectors/index.js
+++ b/injectors/index.js
@@ -37,4 +37,10 @@ try {
     console.log(`Unsupported argument "${process.argv[2]}", exiting..`);
     process.exit(1);
   }
-})().catch(e => console.error('fucky wucky', e));
+})().catch(e => {
+  if (e.code === 'EACCES' && process.platform === 'linux') {
+    console.log('Missing required permissions, rerun with root privileges.');
+  } else {
+    console.error('fucky wucky', e)
+  }
+});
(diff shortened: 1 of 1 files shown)


Weilbyte:
Will make generic for all platforms.

intrnl (changes requested):
As with the other duplicate PR, try to make it generic, and have it be a suggestion.

Something like,
`Unable to inject, you might need sudo or Administrator rights.`

Weilbyte:
Hope that's a bit better.

intrnl (approved):

injectors/index.js
```diff
@@ -37,4 +37,10 @@ try {
     console.log(`Unsupported argument "${process.argv[2]}", exiting..`);
     process.exit(1);
   }
-})().catch(e => console.error('fucky wucky', e));
+})().catch(e => {
+  if (e.code === 'EACCES') {
+    console.log(`Unable to ${process.argv[2]} because of missing permissions. Consider rerunning with elevated rights. `);
```
cyyynthia: Trailing space
```suggestion
    console.log(`Unable to ${process.argv[2]} because of missing permissions. Consider rerunning with elevated rights.`);
```

cyyynthia (changes requested):

injectors/index.js
Weilbyte: Fixed.

cyyynthia (approved):

Issue #352, opened 2020-07-11
Linux installer doesn't like non-standard discord paths


oatmealine:
**Describe the bug**
When installing on Linux with a non-standard Discord path (eg. /usr/share/discordcanary/ instead of /usr/share/discord-canary) the installer doesn't give an option to select your own path or even a descriptive error, and instead returns

`TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined`

The process detection has to fail for the installer to try to use the legacy method, but i'm not sure what exactly made it fail on my end

**To Reproduce**

1. Close Discord
2. Move Discord from whatever directory its installed to to a different path
3. Try to install Powercord

**Expected behavior**
Instead of giving an error, the installer should let the user enter their own path when a valid one isn't found, or at least handle Discord not being found and tell the user an error saying so.

**Desktop (please complete the following information):**
 - OS: Solus x86_64, Linux 5.6.18-156.current
 - Powercord version: Doesn't apply

**Additional context**
 - [x] I checked if Discord and/or Powercord are up to date
 - [x] I made sure this is not related to an external plugin or a theme

Weilbyte:
Opened a PR for this. Hopefully, they'll take a look :)

PR #356, opened 2020-07-15
Prompt for custom install path 

Weilbyte:
While both Windows and macOS have standard/consistent install locations for the Discord Canary client, Linux doesn't. The script that gets the install path for Linux looks through common paths but If the user doesn't have it installed in those locations then it returns `undefined` and errors (see #352).

This change checks if the path is `undefined` and if it is, it prompts the user to enter the path of their actual install location.

Not sure if asking for input here is right but if not I can just remove that bit.

### DIFF
--- a/injectors/linux.js
+++ b/injectors/linux.js
@@ -1,6 +1,9 @@
 const { join } = require('path');
 const { existsSync } = require('fs');
 const { execSync } = require('child_process');
+const readline = require('readline');
+
+
 
 exports.getAppDir = async () => {
   const discordProcess = execSync('ps x')
@@ -17,7 +20,27 @@ exports.getAppDir = async () => {
       '/opt/discord-canary',
       '/opt/DiscordCanary'
     ];
-    const discordPath = paths.find(path => existsSync(path));
+    let discordPath = paths.find(path => existsSync(path));
+
+    if (discordPath === undefined) {
+      const readlineInterface = readline.createInterface({
+        input: process.stdin,
+        output: process.stdout
+      });
+
+      console.log('A Discord Canary installation sdcould not be found at the usual paths.')
+       discordPath = await new Promise(resolve => readlineInterface.question('Provide your Discord Canary install location: ', customDiscordPath => {
+        if (existsSync(customDiscordPath)) {
+          readlineInterface.close();
+          resolve(customDiscordPath);
+        } else {
+          console.log('Path provided is invalid.');
+          process.exit(1);
+        }
+        
+      }));
+    }
+
     return join(discordPath, 'resources', 'app');
   }
 
--- a/injectors/linux.js
+++ b/injectors/linux.js
@@ -3,8 +3,6 @@ const { existsSync } = require('fs');
 const { execSync } = require('child_process');
 const readline = require('readline');
 
-
-
 exports.getAppDir = async () => {
   const discordProcess = execSync('ps x')
     .toString()
--- a/injectors/linux.js
+++ b/injectors/linux.js
@@ -28,13 +28,13 @@ exports.getAppDir = async () => {
         output: process.stdout
       });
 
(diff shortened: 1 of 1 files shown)


injectors/linux.js
```diff
@@ -17,7 +20,27 @@ exports.getAppDir = async () => {
...
-    const discordPath = paths.find(path => existsSync(path));
+    let discordPath = paths.find(path => existsSync(path));
+
+    if (discordPath === undefined) {
+      const readlineInterface = readline.createInterface({
+        input: process.stdin,
+        output: process.stdout
+      });
+
+      console.log('A Discord Canary installation sdcould not be found at the usual paths.')
```
LayneTheEpic, line 31: Typo on "could"

injectors/linux.js
```diff
@@ -17,7 +20,27 @@ exports.getAppDir = async () => {
...
+       discordPath = await new Promise(resolve => readlineInterface.question('Provide your Discord Canary install location: ', customDiscordPath => {
+        if (existsSync(customDiscordPath)) {
+          readlineInterface.close();
+          resolve(customDiscordPath);
+        } else {
+          console.log('Path provided is invalid.');
```
LayneTheEpic, line 37: Should be more indicative that the installer is now going to close, instead of suddenly stopping

injectors/linux.js
LayneTheEpic, line 31: Typo on 'could'

LayneTheEpic, line 31:
```suggestion
      console.log('A Discord Canary installation could not be found at the usual paths.')
```
Typo on 'could'

injectors/linux.js
```diff
@@ -17,7 +20,27 @@ exports.getAppDir = async () => {
...
+        input: process.stdin,
+        output: process.stdout
+      });
+
+      console.log('A Discord Canary installation sdcould not be found at the usual paths.')
+       discordPath = await new Promise(resolve => readlineInterface.question('Provide your Discord Canary install location: ', customDiscordPath => {
+        if (existsSync(customDiscordPath)) {
+          readlineInterface.close();
+          resolve(customDiscordPath);
+        } else {
```
LayneTheEpic, line 36: Should be more indicative that the process will stop, or be better handled

injectors/linux.js
LayneTheEpic, line 37: Should be better indicated that the process will stop, or handled otherwise

injectors/linux.js
lanye74:
```suggestion
      console.log('A Discord Canary installation could not be found at the usual paths.')
```
Typo on 'could'

injectors/linux.js
```diff
@@ -17,7 +20,27 @@ exports.getAppDir = async () => {
...
+          console.log('Path provided is invalid.');
+          process.exit(1);
+        }
```
lanye74, line 37: Should be better indicated that the process will stop, or handled otherwise

lanye74 (changes requested):
Please don't hurt me for all the notifications I'm bad at reviews xd

injectors/linux.js
Weilbyte: Thanks, will fix it.

injectors/linux.js
Weilbyte: Mentioned aborting.

injectors/linux.js
```diff
@@ -1,6 +1,9 @@
 const { join } = require('path');
 const { existsSync } = require('fs');
 const { execSync } = require('child_process');
+const readline = require('readline');
+
```
cyyynthia, line 5: why adding those empty lines?

Weilbyte: Oop didn't notice, fixed that.

cyyynthia (approved):

PR #364, opened 2020-07-21
Added search command

ripSquid:
I added a search command to be able to search for a song and play it automatically inside discord.

I am unsure if powercord spotify permissions are needed. So just add a change if needed.


src/Powercord/plugins/pc-spotify/commands/search.js
```diff
@@ -0,0 +1,38 @@
+const SpotifyAPI = require('../SpotifyAPI');
+
+module.exports = {
+    command: 'search',
```
cyyynthia: This command is already in use by [another plugin](https://github.com/Juby210/quick-search), which would cause a conflict

src/Powercord/plugins/pc-spotify/commands/search.js
```diff
@@ -0,0 +1,38 @@
...
+
+        SpotifyAPI.search(args.join(" "), 'track', 1).then((body) => {
+            let tracksArray = body['tracks']['items']
+            if (tracksArray.length > 0) {
+                let trackURL = tracksArray[0]['uri']
+                return SpotifyAPI.play({
+                    'uris': [trackURL]
+                })
+            } else {
+                return {
```
cyyynthia: This return will never be seen by our command handler since it is inside a promise

cyyynthia (changes requested):
There's a few things that need attention, and there needs to be some changes regarding linting as you can see from the [Travis build logs](https://travis-ci.org/github/powercord-org/powercord/builds/710493973#L227)

ripSquid:
I think I have fixed the linting issues and also one forgotten import. I think I have also fixed the return. Someone suggested the command "find" so if that is available I'm all good.

src/Powercord/plugins/pc-spotify/commands/search.js
ripSquid: Renamed to find

src/Powercord/plugins/pc-spotify/commands/search.js
ripSquid: Also returned the promise
`return SpotifyAPI.search(args.join(' '), 'track', 1).then((body) => {`

cyyynthia (approved):

Issue #332, opened 2020-05-31
Uninstall plugin button unimplemented


lanye74:
**Describe the bug**
In settings, when you go to uninstall a plugin, I get the error:
```
Uncaught TypeError: this.onUninstall is not a function
    at onClick (C:\Users\layne\OneDrive\Documents\powercord\src\Powercord\plugins\pc-moduleManager\components\parts\BaseProduct.jsx:62:35)
    at Object.h (https://canary.discordapp.com/assets/4afc54d4d5371f70edf8.js:88:1026)
    at S (https://canary.discordapp.com/assets/4afc54d4d5371f70edf8.js:88:1169)
    at https://canary.discordapp.com/assets/4afc54d4d5371f70edf8.js:88:1315`...
```

**To Reproduce**

**Expected behavior**
Some sort of uninstall modal should appear to confirm that I want to remove the plugin, I guess ;P

**Desktop (please complete the following information):**
 - OS: Windows
 - Powercord version: Commit `4ec1e3a`

**Additional context**
 - [x] I checked if Discord and/or Powercord are up to date
 - [x] I made sure this is not related to an external plugin or a theme (not due to themes; no extra plugins)
 
Relevant file: `BaseProduct.jsx`; line 62
Relevant code:
```js
  renderFooter () {
    if (!this.props.product.discord && typeof this.props.goToSettings !== 'function' && typeof this.props.onUninstall !== 'function') {
      return null;
    }

    return (
      React.createElement(React.Fragment, null
        , React.createElement(Divider, {__self: this, __source: {fileName: _jsxFileName, lineNumber: 45}})
        , React.createElement('div', { className: "powercord-product-footer", __self: this, __source: {fileName: _jsxFileName, lineNumber: 46}}
          , this.props.product.discord && // @todo: i18n
          React.createElement(Tooltip, { text: "Go to their Discord support server"     , __self: this, __source: {fileName: _jsxFileName, lineNumber: 48}}
            , React.createElement(Clickable, { onClick: () => this.goToDiscord(this.props.product.discord), __self: this, __source: {fileName: _jsxFileName, lineNumber: 49}}
              , React.createElement(Discord, {__self: this, __source: {fileName: _jsxFileName, lineNumber: 50}})
            )
          )
          , typeof this.props.goToSettings === 'function' && // @todo: i18n
          React.createElement(Tooltip, { text: "Settings", __self: this, __source: {fileName: _jsxFileName, lineNumber: 54}}
[...]

PR #369, opened 2020-07-28
how did you even manage to accidentally do this

CatboyCoding:


CatboyCoding:
like for real

cyyynthia:
Please avoid unnecessary comments on the repo, as people get junk email thanks to this.

Moreover, the PR is not targeted to the right branch and this has been left intentionally as-is due to the lack of check for whether the plugin in internal or not (and don't allow removal if internal). There's a reason why it's tracked as "unimplemented" in #332

PR #370, opened 2020-07-28
Added .local path

dr03m:
Used for injecting Powercord into Canary that was installed into the user's home directory, such as people who used this script.
https://gist.github.com/lnoering/e914013e2b8699327ef44b62862e9ac1

### DIFF
--- a/injectors/linux.js
+++ b/injectors/linux.js
@@ -15,7 +15,8 @@ exports.getAppDir = async () => {
       '/usr/share/discord-canary',
       '/usr/lib64/discord-canary',
       '/opt/discord-canary',
-      '/opt/DiscordCanary'
+      '/opt/DiscordCanary',
+      '~/.local/bin/DiscordCanary/'
     ];
     const discordPath = paths.find(path => existsSync(path));
     return join(discordPath, 'resources', 'app');


injectors/linux.js
```diff
@@ -15,7 +15,8 @@ exports.getAppDir = async () => {
       '/usr/share/discord-canary',
       '/usr/lib64/discord-canary',
       '/opt/discord-canary',
-      '/opt/DiscordCanary'
+      '/opt/DiscordCanary',
+      '~/.local/bin/DiscordCanary/'
```
cyyynthia, line 19: I'm unsure if  `~` is safe to use here. Did it cause any troubles on your end?

Bowser65:
I went ahead and implemented it in 31f4f0b914110e170541a4159b78825a29f749f9 with an implementation that resolves the homedir is a more robust way (even when ran as sudo it'll resolve to the current user's home dir). Since it's local I guess require('os').homedir() could've been enough, but that's just to be sure since our installation guide tells Linux users to use sudo.

Thanks for pointing out this install path :+1:

dr03m:
Ah, thanks. I believe doing `~` will do the job just fine, too.

Bowser65:
Yeah I'm unsure about node's handling of this, I feel it's safer to have it resolve an absolute path

intrnl:
Does `~` actually resolve to the home dir? I don't think it does...

PR #372, opened 2020-07-31
crx url fix (react devtools)

CatboyCoding:
very nice


CatboyCoding:
crxToZip apparently does **not** support crx3. Closing this while I find another fix.

PR #379, opened 2020-08-14
Added unsupported plugin warning

TheArcaneBrony:
Detects wether a .git folder exists, or a .ignore-no-git file (generated by the Ignore button)
Has a button to create a file so the warning doesn't show up in the client
I hope that hereby we can assist users that use the Github download button without them needing to ask for support

### DIFF
--- a/src/Powercord/managers/plugins.js
+++ b/src/Powercord/managers/plugins.js
@@ -1,34 +1,36 @@
 const { resolve } = require('path');
 const { readdirSync } = require('fs');
 const { rmdirRf } = require('powercord/util');
+var shell = require('electron').shell;
+var fs = require('fs');
 
 module.exports = class PluginManager {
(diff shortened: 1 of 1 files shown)


TheArcaneBrony:
Also possible: make it opt in (for example by checking for a .gitignore file, or a manually added .release file)

cyyynthia:
As mentioned in the powercord-core channel (before the nuke) before the PR was made, I'm still not sure if we should have such kind of notifications.

But anyway, v2 being officially in maintenance only I'll close this. I'll think about it for v3, but again, I'm unsure about that.

TheArcaneBrony:
It does indeed not make sense for v2, once v3 comes around (and maybe of we have a plugin store) i'll see how fitting it would be. Doesn't really make sense without the store as ypu mentioned.

PR #384, opened 2020-08-30
Switch from system-font-families to font-list

cillynder:
`system-font-families` depends on `ttfinfo` which has been unmaintained for over 7 years.

`ttfinfo` loads into memory **all** the content of **every** font file present on a system, even though Powercord (to my knowledge) only uses the font names. This can cause huge startup memory consumption depending on the total filesize of the system's installed fonts.

This PR attempts to fix this by switching from `system-font-families` to `font-list`. `font-list` takes a different approach by calling the command `fc-list` provided by `fontconfig`. The two packages do produce different results. `font-list` outputs less fonts, but the missing fonts are usually some variants of a parent font.

[This file](https://github.com/powercord-org/powercord/files/5146384/system-fonts.txt) lists the fonts installed on my Arch Linux system.
These are text files containing the results each package produced on my personal machine.
[font-list.txt](https://github.com/powercord-org/powercord/files/5146381/font-list.txt)
[system-font-families.txt](https://github.com/powercord-org/powercord/files/5146382/system-font-families.txt)

Note that what I have said above about mainly targets Linux. I have not tested this on a Windows or macOS system and do not know of this PR's effectiveness on said systems; however, I believe it shouldn't be causing any huge regressions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,12 +19,12 @@
   },
   "dependencies": {
     "codemirror": "^5.54.0",
+    "font-list": "^1.2.11",
     "less": "^3.11.1",
     "node-watch": "^0.6.4",
     "sass": "^1.26.7",
     "stylus": "^0.54.7",
     "sucrase": "^3.15.0",
-    "system-font-families": "^0.4.1",
     "unzip-crx": "^0.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


cyyynthia:
I remember trying this while developing it but I ended up throwing it away for some reason I don't remember. I'll investigate this later today.

cyyynthia:
I haven't got time to look at this closely; until I can I commented out the faulty part in 204d21cc830ca01d1b09bc1c2369eb8f73a9a095 so the memory spike should disapear.

cyyynthia:
Due to v2 being maintenance only now and this PR targeting code that'll never see production stage in its current state, I'll close the PR. Thanks for bringing this to my attention though, I'll definitely take that into account for v3.

PR #394, opened 2020-09-21
fix announcements

redstonekasi:
got everything except the `facebook` and `survey` colors, so i removed them for now.


cyyynthia (approved):
They probably got removed, not a big deal

PR #416, opened 2020-10-19
made travis happy by adding semicolons

the-ashh:
~~the little red x was bugging me~~


cyyynthia:
I appreciate the attention, but since v2 is being the main upstream branch I won't merge such small change since it'll trigger an unnecessary update to all users.

It'll eventually get fixed in a larger push, when fixing something for example. Thanks though!

PR #419, opened 2020-10-24
Node-watch upgrade, plus stopped ESLint error.

SoffiMeower:


cyyynthia:
I won't merge this for the same reason as #416. There is no reason to upgrade node-watch and this would trigger an update on all Powercord users which is not something we want to do for something that minor.

PR #422, opened 2020-10-31
prevent updater crashes when commits[0] is undefined

AAGaming00:
prevent the rare case of the updater getting stuck on the last item


cyyynthia:
Merged in f25483c

PR #430, opened 2020-11-25
Fixed pc-tags .list command not having proper grammar when values are 0

dutake:


cyyynthia:
Seems like my attempt at merging it locally went wrong heh, merged in 5c4eb09

PR #431, opened 2020-11-25
"Fix" pc-docs

valentinegb:
**And by "fix" I mean I shoved in a bunch of random code and deleted stuff my small brain couldn't understand.** The result of this isn't too bad though, the documentation page is functional now. I am aware that this is entirely pointless, because the documentation currently available is barely valid and is being rewritten, and I don't care. I honestly just think it looks nice. There's not much to lose anyway, since I'm assuming `pc-docs` is going to be rewritten in `v3`. If you think changes need to be made, do it, my code is awful.


cyyynthia:
Powercord Labs has been dropped which means docs locked out (unless you run a snippet in devtools ig) which makes them completely unsupported.

The docs on api-v2 are close to useless, and the pc/docs is actually more useful since it actually tells you where to look. There also a lot of codestyle issues, and adding marked as a dependency is really not ideal (it's not even placed in the right spot).

Issue #436, opened 2020-12-11
Emote Utility context menu is wrongly spaced


sapphonie:
**Describe the bug**
Emote utility has odd spacing with "Create in" and "Clone in" context menus

**To Reproduce**

Click a picture or emote to create an emote out of or clone
Odd spacing occurs:

https://i.imgur.com/HuCQh8i.png

**Expected behavior**
Context menu should look something like this, in following the style of every other discord context menu item:

https://i.imgur.com/TknYJ3P.png

**Desktop (please complete the following information):**
 - OS: Win 10 Pro x64
 - Powercord version: Official 7813234 v2

**Additional context**
 - [yup] I checked if Discord and/or Powercord are up to date
 - [yup] I made sure this is not related to an external plugin or a theme

cyyynthia:
This is the intended behavior. The words aren't meant to be read together.

sapphonie:
Why is this intended? It does not match the style of literally any other context menu I've seen, in native discord or a plugin.

It doesn't matter if the words aren't "meant to be" read together, they still are by most english speakers.

ghost:
the blank space is probably referring to the emote you're planning to clone. "Clone (the emote you right clicked) to"

Rexogamer:
> This is the intended behavior. The words aren't meant to be read together.

Wouldn't it be better to have something here suggesting what you're creating instead of a blank space? I can definitely see the confusion if you don't know this is intended.

PR #437, opened 2020-12-11
update emote submenu to be properly spaced

sapphonie:
solves #436


cyyynthia:
This is the intended behavior.

PR #438, opened 2020-12-24
theme manager & fix uninstall buttons

apart:


cyyynthia:
For people stumbling across this PR, the PR author is someone who have been banned from interacting with the organization (and circumvented this ban using an alternate account). We are not accepting any contribution from people who repeatedly distributed malware, period.

Powercord v2 is also in maintenance-only since a while now, which means we won't accept new features to be merged.

PR #442, opened 2021-01-09
Handle error while creating app directory

max-niederman:
I had an issue injecting Powercord into [the `discord-canary-electron-bin` AUR package](https://aur.archlinux.org/packages/discord-canary-electron-bin/). After some troubleshooting, I found that the problem was a discrepancy in the way that this specific package installed Discord Canary.

This adds a check which informs the user that their Discord install is unusual.

### DIFF
--- a/injectors/main.js
+++ b/injectors/main.js
@@ -23,7 +23,14 @@ exports.inject = async ({ getAppDir }) => {
     return false;
   }
 
-  await mkdir(appDir);
+  try {
+    await mkdir(appDir);
+  } catch (err) {
+    console.log('Your Discord Canary install appears to be broken.', '\n');
+    console.log(`${AnsiEscapes.YELLOW}NOTE:${AnsiEscapes.RESET} This issue has been known to happen using non-official Discord Canary installs. Try using the official installation method if you're not already.`);
+
+    return false;
+  }
   await Promise.all([
     writeFile(
       join(appDir, 'index.js'),


cyyynthia:
Your PR is completely breaking the code style of the project. I know eslint is badly configured and doesn't cover this folder, but when contributing to a project you really shouldn't mess around with the formatting (even if the editor does it "automatically").

max-niederman:
@cyyynthia I actually didn't even notice that when I first made this PR (as you guessed, my editor did it). It's fixed now.

Kodehawa:
"I found that the problem was a discrepancy in the way that this specific package installed Discord Canary."

As the packager: this is so it uses the system electron, instead of the bundled electron by discord (which works like ass on linux). There must be a way to make Powercord detect it instead of making it so it crashes (and I think it detects the sibling, discord_arch_electron)

Vendicated:
there is indeed a very simple way to make it work:
```diff
diff --git a/injectors/linux.js b/injectors/linux.js
index d3086a4..68b4e60 100644
--- a/injectors/linux.js
+++ b/injectors/linux.js
@@ -14,6 +14,7 @@ const { BasicMessages, AnsiEscapes } = require('./log');
 const homedir = execSync('grep $(logname) /etc/passwd | cut -d ":" -f6').toString().trim();
 
 const KnownLinuxPaths = Object.freeze([
+  '/usr/lib/discord-canary',
   '/usr/share/discord-canary',
   '/usr/lib64/discord-canary',
   '/opt/discord-canary',
@@ -51,10 +52,10 @@ exports.getAppDir = async () => {
       }
     }
 
-    return join(discordPath, 'resources', 'app');
+    return join(discordPath, 'app');
   }
 
   const discordPath = discordProcess[4].split('/');
   discordPath.splice(discordPath.length - 1, 1);
-  return join('/', ...discordPath, 'resources', 'app');
+  return join('/', ...discordPath, 'app');
 };
```

Additionally, you must copy `/usr/bin/discord-canary` to `/usr/local/bin/discord-canary` and make it launch `/usr/lib/discord-canary/app` instead of `/usr/lib/discord-canary/app.asar`


(obviously this patch would break injections for all normal discord installs, but that could be solved with a simple if statement)

AkechiShiro:
Is there any plan on merging this PR someday, or are there too many issues with it at the moment ? (Or maybe there are other more important PRs)

huantianad:
@AkechiShiro It might be better to patch the injector in the `powercord-git` AUR package instead of doing this upstream?

AkechiShiro:
You're right, I'm guessing to do such a patch, following this course of action is right ?

The changes to do on the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=powercord-git) are probably add two `sed` inline commands that changes the two lines mentioned by @Vendicated inside `${srcdir}/injectors/linux.js`, copying with cp `/usr/bin/discord-canary` to `/usr/local/bin/discord-canary` and then doing a last `sed` inline to replace `/usr/lib/discord-canary/app` instead of `/usr/lib/discord-canary/app.asar` inside `/usr/local/bin/discord-canary`.

Edit: I actually noticed I missed one more line that needs to be added inside the diff.

AkechiShiro:
Having tried the patch proposed, it doesn't look like it worked for me for some reason.

Running the modified `discord-canary` yielded this error :
```
Error launching app
Unable to find Electron app at /usr/lib/discord-canary/app

Cannot find module '/usr/lib/discord-canary/app'
Require stack:
- /usr/lib/electron/resources/default_app.asar/main.js
-
```

If anyone has a clue on how to fix this.

max-niederman:
This PR doesn't actually *fix* the issue at hand, it just detects the problem and fails gracefully. I think it would be best if discussion on a real solution was in its own issue, ideally with this PR merged in the meantime.

huantianad:
@AkechiShiro Make sure you run `sudo powercord inject` first, then try to launch.

Mirivel:
Genial

PR #445, opened 2021-01-11
Escape path when injecting

magistau:
A better way to do #26 IMHO.
#26 allows `'` in path, but disallows \` and `${`. This PR allows all of them and `\` by escaping `'` and `\`. 
Should not break Windows compatibility since the escaping is made after the separator replacement, but needs some testing.

### DIFF
--- a/injectors/main.js
+++ b/injectors/main.js
@@ -27,7 +27,10 @@ exports.inject = async ({ getAppDir }) => {
   await Promise.all([
     writeFile(
       join(appDir, 'index.js'),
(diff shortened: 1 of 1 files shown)


cyyynthia:
Is there any real benefit in doing it that way? In normal circumstances Discord Canary's installation path shouldn't include those characters.

magistau:
> In normal circumstances Discord Canary's installation path shouldn't include those characters.

Is it more likely to include `'`?

AstroOrbis (approved):

cyyynthia:
There is no real advantage over changing how the path is written; closing this

PR #453, opened 2021-01-28
Fixed settings path and added window frame option

zt64:
This fixes a bug that caused the try catch statement in browserWindow.js to always fail as a result from the require path going back one directory too far. It also adds an option to the advanced settings that allows toggling of the window frame. I've only added English translations because I do not have any other language skill and using an online translator isn't the best way.

### DIFF
--- a/src/browserWindow.js
+++ b/src/browserWindow.js
@@ -9,12 +9,14 @@ const { BrowserWindow } = require('electron');
 
 let settings = {};
 let transparency = false;
+let frame = true;
 let ewp = false;
 try {
-  settings = require(join(__dirname, '../../settings/pc-general.json'));
+  settings = require(join(__dirname, '../settings/pc-general.json'));
   transparency = settings.transparentWindow;
+  frame = settings.windowFrame;
   ewp = settings.experimentalWebPlatform;
-} catch (e) {}
+} catch (e) { console.error(e) }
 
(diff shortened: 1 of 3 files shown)


ShinzenATT:
I tested your forked repository but it doesn't seem to work with linux properly where it results in a blank windows. But the directory fix in src/browserWindows.js fixes the transparent windows feature.

zt64:
> I tested your forked repository but it doesn't seem to work with linux properly where it results in a blank windows. But the directory fix in src/browserWindows.js fixes the transparent windows feature.

The repository isn't really meant to be used, I just modified the files of my local install on the v2 branch.

src/browserWindow.js
```diff
@@ -9,12 +9,14 @@ const { BrowserWindow } = require('electron');
...
+let frame = true;
 let ewp = false;
 try {
-  settings = require(join(__dirname, '../../settings/pc-general.json'));
+  settings = require(join(__dirname, '../settings/pc-general.json'));
   transparency = settings.transparentWindow;
+  frame = settings.windowFrame;
   ewp = settings.experimentalWebPlatform;
-} catch (e) {}
+} catch (e) { console.error(e) }
```
cyyynthia: Failure here is to be expected if the user doesn't have a config file (e.g. on a clean install) which is why we ignore it silently.

PR #454, opened 2021-01-28
remove cynthia's commits from today as they break the injector

braddevans:
the said errors from cyyynthia's commits from today and yesterday

as of `58811c935a3f5cccf691c470b88ca6a3d00456ce`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,8 +1,0 @@
-*.template.js
-/src/Powercord/plugins/*
-!/src/Powercord/plugins/pc-*
-
-# tmp
-/src/core.old/*
-/.v2/src/Powercord/plugins/*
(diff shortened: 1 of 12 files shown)


Oocrop:
screenshots you provided suggest that this is actually a problem with one specific plugin (https://github.com/HooferDevelops/Custom-User-Tags)
consider creating issue at that plugin's repo

cyyynthia:
The commits are not related to the injector issues. They are **required** in order for Discord to actually work (otherwise you end up with a browser-ified Discord client).

Make sure your Discord client is up to date, and as mentioned please reach out to the plugin developer of Custom user Tags since, as the screenshot tells, is from where the error comes from.

PR #455, opened 2021-01-29
Powercord plugins separate from user plugins 

userXinos:
Separate category for Powercord plugins with a button that activates them all

_I'm the only one bothered that my plugins are mixed with plugins that I did not install? 
and somehow I don't want to delete them..._

Prews:
one
img1
two
img2


LoganDark:
Personally I would prefer a second tab rather than a collapse, but this still totally works.

src/Powercord/plugins/pc-moduleManager/components/manage/Base.jsx
```diff
@@ -79,6 +80,48 @@ class Base extends React.Component {
...
+        {items.map(item => this.renderItem(item))}
+      </Category>
+    );
+  }
+
+  renderEnableAllButton (items) {
+    const needEnable = items.filter(p => !p.ready);
+    return (
+      <ButtonItem
+        button={(needEnable.length) ? Messages.POWERCORD_PLUGINS_ENABLE_ALL_PC_GO : Messages.POWERCORD_PLUGINS_ALL_ENABLED}
```
cyyynthia: You could just make a simple button without a label in front, and put the label in it. I don't think it's even worth checking if everything is enabled and disable it if it's the case (avoids having to iterate over the plugin array).

cyyynthia (changes requested):
I don't fully agree with how the logic is implemented (it's honestly a bit inefficient on how it iterates over the array) but oh well not like v2 is here to stay for long (hopefully), so it's good enough. The button "Go" really looks off though.

src/Powercord/plugins/pc-moduleManager/components/manage/Base.jsx
userXinos: well, if forced enable does not cause problems...

PR #456, opened 2021-01-29
Make the Spotify Plugin function again

bakzkndd:
removed the websocket and replaced it with an interval through the api

- some code might be irrelevant now but i didn't want to accidentally break it again

### DIFF
--- a/src/Powercord/plugins/pc-spotify/index.js
+++ b/src/Powercord/plugins/pc-spotify/index.js
@@ -38,14 +38,12 @@ class Spotify extends Plugin {
   }
 
   startPlugin () {
-    return; // Please shut the f up
 
     /* eslint-disable */
     this.loadStylesheet('style.scss');
     this._injectModal();
     this._patchAutoPause();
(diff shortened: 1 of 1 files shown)


cyyynthia:
The problem with the plugin is not limited to the WebSocket issues but with the plugins randomly causing hard crashes for some people, which is not something this PR addresses at all. We cannot re-enable the plugin as-is with a chance of making some clients unusable.

Polling is also extremely inefficient and something that should be avoided as much as possible (especially since in our case there are alternative, better ways of doing things). Sure, it's *can* do the job as a temporary fix, but as I mentioned there are more severe issues that are not addressed anyway, and on top of that some of the plugin features requires events from the WebSocket.

H4kt:
If plugin causes random crashes for **some** users, then just disable it by default and leave users an option to enable it back. Cuz I had no problems using it and therefore I rly miss it

cyyynthia:
A proper fix has been implemented in 14d8f0b119fee775821792f3f148eb0641b04e79

PR #464, opened 2021-02-24
Invite command

ghost:
Added a new command that sends an invite to the Powercord Discord server in the chat you are

### DIFF
--- a/src/Powercord/plugins/pc-commands/commands/invite.js
+++ b/src/Powercord/plugins/pc-commands/commands/invite.js
@@ -1,0 +1,10 @@
+module.exports = {
+    command: 'invite',
+    description: 'Sends an invite to the Powercord discord server in the chat you are in',
+    usage: '{c}',
+    showTyping: true,
+    executor: (args) => ({
+      send: true,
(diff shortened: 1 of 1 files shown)


NurMarvin:
Sounds pretty useless to me. Would probably also result in accidental server advertising.
Also, what's so hard about just doing this?
image

zt64:
I agree with @NurMarvin, it would probably be better as a separate plugin.

ghost:
my pc shut down while writing this bruh

I mean it's just to have it handy, it's not something SUUUPER useful, but sometimes it helps
About the accidental server advertising, you're right, maybe I can make it a client-side message with a warning related to that
And about the invite thing, as I said, is just to have it handy

PD: Not entirely sure why the check failed

leKamikaze1:
wait, why did this get closed? reopening

ghost:
wait, why did this get closed? reopening

DracoClaw:
If you really need the invite that often, just create a tag for it.

ghost:
I didn't even know that command existed. Well, guess the invite command is useless then, thanks anyways

PR #475, opened 2021-04-05
Filter plugins and themes based on if they are enabled or not

RazerMoon:
Previews:


environment-dev (approved):
lgtm, i'd like to see this added

PR #481, opened 2021-04-07
Reload command

Vendicated:
Adds a reload command to pc-moduleManager which remounts the specified plugin. This is extremly useful for development as you can now reload plugins without having to restart discord. Additionally, the autocomplete of enable and disable is now case insensitive

### DIFF
--- a/src/Powercord/plugins/pc-moduleManager/commands/enable.js
+++ b/src/Powercord/plugins/pc-moduleManager/commands/enable.js
@@ -35,7 +35,7 @@ module.exports = {
 
     return {
       commands: plugins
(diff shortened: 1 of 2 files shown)


AstroOrbis (approved):

mic0ishere:
Why wasn't it pushed yet? It is simplifies the job of developing plugins so much…

cyyynthia:
Because we are not interested in adding new features at this time. We only keep Powercord working with its current set of features while we work on v3. This is just a command-ified way of using `powercord.pluginManager.remount` in DevTools.

PR #485, opened 2021-04-20
Add flatpak support

Vendicated:
Adds flatpak paths and prints permission override note for flatpak users

___

#### Additional note for flatpak users finding this PR (might want to add this to README in case this gets merged? idk)

Due to flatpaks being sandboxed, powercord doesn't have access to the git binary which breaks auto updating. Assuming your powercord folder is in `~/.local/share/powercord`, you can use this simple script to update all your plugins manually
```bash
#!/bin/bash
shopt -s globstar

for folder in ~/.local/share/powercord/**/.git; do
  (cd "$folder/.." && echo "Pulling $PWD" && git pull)
done
```

### DIFF
--- a/injectors/linux.js
+++ b/injectors/linux.js
@@ -18,7 +18,9 @@ const KnownLinuxPaths = Object.freeze([
   '/usr/lib64/discord-canary',
   '/opt/discord-canary',
   '/opt/DiscordCanary',
-  `${homedir}/.local/bin/DiscordCanary` // https://github.com/powercord-org/powercord/pull/370
+  `${homedir}/.local/bin/DiscordCanary`, // https://github.com/powercord-org/powercord/pull/370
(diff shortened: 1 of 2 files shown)


AstroOrbis (approved):

Lunarequest:
a suggestion as someone who has worked on flatpaks with git is to extract the prebuilt bins from a sdk. For example in the github  desktop flatpak we do this to get git, it shouldn't be hard to replicate this by grabbing the required stuff from the Sdk, I would suggest getting it from `org.freedesktop.Sdk`
```yaml
- name: git
    buildsystem: simple
    build-commands:
      - install -Dm0755 $(which git) /app/bin/
      - install -Dm0755 $(which git-cvsserver) /app/bin/
      - install -Dm0755 $(which git-receive-pack) /app/bin/
      - install -Dm0755 $(which git-shell) /app/bin/
      - install -Dm0755 $(which git-upload-archive) /app/bin/
      - install -Dm0755 $(which git-upload-pack) /app/bin/
 ```

PR #499, opened 2021-05-10
Preinstall basic plugins

SomeAspy:
I probably am stupid for making this PR, but whats the worst that could happen?


cyyynthia:
The plugins do not match the naming scheme requirement, they have bugs that haven't been fixed that can lead to crashes in some circumstances, we've been refusing to add **any** new plugin to the core for years and Powercord v2 is maintenance only.

SomeAspy:
fair enough.

PR #502, opened 2021-05-27
Custom Badges on new UserPopouts

notsapinho:


NurMarvin (changes requested):
This no longer seems to work on my end, client crashes with
```
TypeError: Cannot read property '0' of undefined
    at res.props.children.props.children.<computed>.type (VM929 {path to powercord}\powercord\src\Powercord\coremods\badges\index.js:86)
```

src/Powercord/coremods/badges/index.js
```diff
@@ -58,6 +70,78 @@ async function injectUsers () {
...
+    }
+  });
+
+  inject('pc-badges-popout-render', UserPopOut.prototype, 'render', function (_, res) {
+    const renderer = res.props.children.props.children[0].type;
+
+    res.props.children.props.children[0].type = (props) => {
+      const res = renderer(props);
+
+      const renderer2 = res.props.children[1].props.children[0].type;
```
NurMarvin: This part seems to be broken

cyyynthia:
This has been implemented as a side effect of the fix in 224b396db3ea8b9474d80fc32919d40f4be01a88

PR #540, opened 2021-08-26
case insensitive autocomplete

fawni:
this makes the enable and disable plugin commands case insensitive.
scrot


ConnorDoesDev:
looks good

PR #543, opened 2021-08-27
adding a update script....

Blazzycrafter:
But its no restart discord only a kill

### DIFF
--- a/update.bat
+++ b/update.bat
@@ -1,0 +1,16 @@
+@echo off
+echo 0: Part ZERO: kill discord...
+start /wait cmd /c taskkill /IM DiscordCanary.exe /f
+echo DONE!
+echo 1: Part ONE: unplug Powercord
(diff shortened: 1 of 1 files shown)


suprstarrd:
This really isn't needed, Powercord has updates built-in

Blazzycrafter:
yeah and discord broke againg one plugin and we CAN´T use the bultin updater -_-

pythonmcpi:
IMO this script is unnecessary. Just run `cd powercord && npm run unplug && git pull && npm i && npm run plug` and you're good to go. Also, powercord is supposed to be cross platform, and this script won't work on Mac or Linux. Also, if we really want an auto update script, it's probably better to make it runnable using the npm command for consistency. (So it would be `npm run update`.)

Blazzycrafter:
yeah that's an idea 💡 but I must lern it...
CLOSED

PR #544, opened 2021-08-29
Make README fancier

ghost:
I used the .svg hosted in https://powercord.dev website, if it goes down then it's possible that the image won't load on the readme
A solution to this would be to host the file in this repo itself


ghost:
:zere_zoom3:

PR #556, opened 2021-10-08
Fix query selector in pc-codeblocks

flleeppyy:
For whatever reason, chat message elements are now nested by one, so instead of being by themselves, they now have an `li` parent.


GriefMoDz:
Thanks for the pull request. I'm here to inform you that the proposed change was previously implemented under this commit [here](https://github.com/powercord-org/powercord/commit/ab4d0157ca1c6852a6e08cb4316d5426b57e91cf#diff-8ce33e8d3fbad46edb974997bc48e78eb22447f0c29257c992708e9fa22a8f8dR97). For this reason I'll be closing this pull request.

PR #558, opened 2021-10-09
adding npm run update... (not tested in an linux or mac maschine...)

Blazzycrafter:
using one dependencie and one file i have added an command...

kills discord via fkill an crossplatform kill command (NOT TESTED ON LINUX MAC...) LINK: https://github.com/sindresorhus/fkill
npm run unplug
git pull
npm i
npm run plug


package.json
```diff
@@ -1,34 +1,36 @@
...
+  "license": "SEE LICENSE IN LICENSE",
+  "main": "index.js",
+  "scripts": {
+    "test": "jest",
+    "lint": "eslint --ext js --ext jsx src",
+    "plug": "node injectors/index.js inject --no-exit-codes",
+    "unplug": "node injectors/index.js uninject --no-exit-codes",
+    "inject": "node injectors/index.js inject --no-exit-codes",
+    "uninject": "node injectors/index.js uninject --no-exit-codes",
+    "update": "node injectors/fkill.js; npm run unplug && git pull && npm i && npm run plug"
```
Blazzycrafter: ERROR: node injectors/fkill.js;  the  simicolon xD
```suggestion
    "update": "node injectors/fkill.js npm run unplug && git pull && npm i && npm run plug"

```

package.json
```diff
@@ -1,34 +1,37 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "jest",
+    "lint": "eslint --ext js --ext jsx src",
+    "plug": "node injectors/index.js inject --no-exit-codes",
+    "unplug": "node injectors/index.js uninject --no-exit-codes",
+    "inject": "node injectors/index.js inject --no-exit-codes",
+    "uninject": "node injectors/index.js uninject --no-exit-codes",
+    "update": "node injectors/fkill.js npm run unplug && git pull && npm i && npm run plug"
+
```
Blazzycrafter: Remove NL
```suggestion
```

package.json
```diff
@@ -1,34 +1,37 @@
...
+  },
+  "dependencies": {
+    "codemirror": "^5.57.0",
+    "less": "^3.12.2",
+    "node-watch": "^0.6.4",
+    "sass": "^1.35.2",
+    "stylus": "^0.54.8",
+    "sucrase": "^3.15.0",
+    "unzip-crx": "^0.2.0",
+    "fkill":"7.2.1"
```
Blazzycrafter: Formmatting
```suggestion
    "fkill": "7.2.1"
```

Blazzycrafter (commented):
i remove my found "issues"

injectors/fkill.js
```diff
@@ -0,0 +1,5 @@
+const fkill = require('fkill');
+
+(async () => {
+	await fkill("DiscordCanary.exe", {"force": true, "tree": true});
```
pythonmcpi, line 4: This line will not work on Mac or Linux. The Discord Canary binary's filename does not end with .exe on non-Windows platforms. I would recommend changing the target process name based on the user's platform.

Also, the `force` option should be `false`. If `force` is enabled, (at least on MacOS and Linux) the process will be terminated without allowing it to clean up (since a SIGKILL is sent instead of a SIGTERM). See [this Stack Exchange post](https://unix.stackexchange.com/a/281440) for more information on why you shouldn't use SIGKILL.

`silent` should probably be set to `true` as well. If canary was not running, the user will see a 'Process not found.' without any explanations. The rest of the commands in the chain might not run either. (I'm not sure if node sets an error return code for errors in Promises.)

package.json
```diff
@@ -1,34 +1,36 @@
...
+  },
+  "dependencies": {
+    "codemirror": "^5.57.0",
+    "less": "^3.12.2",
+    "node-watch": "^0.6.4",
+    "sass": "^1.35.2",
+    "stylus": "^0.54.8",
+    "sucrase": "^3.15.0",
+    "unzip-crx": "^0.2.0",
+    "fkill": "7.2.1"
```
pythonmcpi, line 29: Weight of the dependencies [before](https://bundlephobia.com/scan-results?packages=codemirror@5.57.0,less@3.12.2,node-watch@0.6.4,sass@1.35.2,stylus@0.54.8,sucrase@3.15.0,unzip-crx@0.2.0&sortMode=size) and [after](https://bundlephobia.com/scan-results?packages=codemirror@5.57.0,less@3.12.2,node-watch@0.6.4,sass@1.35.2,stylus@0.54.8,sucrase@3.15.0,unzip-crx@0.2.0,fkill@7.2.1&sortMode=size) adding `fkill`. The difference is fairly small but considering that this is to support something that many Powercord users will not use, is it really necessary? Are there any better, more lightweight alternatives? (The answer might be no, more research needs to be done on package options.)

pythonmcpi (changes requested):
Currently only works on Windows. See my comment in `injectors/fkill.js`.

package.json
```diff
@@ -1,34 +1,36 @@
...
+  "license": "SEE LICENSE IN LICENSE",
+  "main": "index.js",
+  "scripts": {
+    "test": "jest",
+    "lint": "eslint --ext js --ext jsx src",
+    "plug": "node injectors/index.js inject --no-exit-codes",
+    "unplug": "node injectors/index.js uninject --no-exit-codes",
+    "inject": "node injectors/index.js inject --no-exit-codes",
+    "uninject": "node injectors/index.js uninject --no-exit-codes",
+    "update": "node injectors/fkill.js npm run unplug && git pull && npm i && npm run plug"
```
pythonmcpi, line 14: The `&&` between the `node injectors/fkill.js` and `npm run unplug` seems to be missing.

Blazzycrafter:
reason:
i dont work anymore at this...

PR #563, opened 2021-10-16
Added Ukranian Translation

zxsleebu:
By Sleebu#0090.
Please give me the badge. I will appreciate that!


cyyynthia:
We are currently not accepting new translations - for future reference translations should be done on https://i18n.powercord.dev

PR #570, opened 2021-10-29
Refactored the injection module to be more efficient

An00nymushun:
Most importantly, I removed the part where it scans through all the injections twice, to find all the ones that match the called method.
I also made the debugging code a bit better by only showing hooks that actually ran and added better performance measurements by using perf_hooks.

### DIFF
--- a/src/fake_node_modules/powercord/injector/index.js
+++ b/src/fake_node_modules/powercord/injector/index.js
@@ -1,70 +1,67 @@
+"use strict";
+
+const { performance } = require('perf_hooks');
+
 window.__$$injectorRuns = [];
 window.__$$injectorRecord = false;
 
+const injectionsById = new Map();
+
 const injector = {
-  injections: [],
+  get injections() { return [...injectionsById.values()]; },
 
   /**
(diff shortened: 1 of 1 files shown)


cyyynthia:
Considering v2 is in maintenance only, there's not much interest in adding new features or doing anything much than base maintenance and ensure it still works with Discord updates.

While scanning twice is not really ideal, it's unlikely there'll ever be enough injections for it to have a real, noticeable impact. Moreover, perf_hooks is only precise up to milliseconds, whereas hrtime is outputting nanoseconds which is very useful for cases where we have sub-millisecond runtimes (which can easily happen with small injections on high end machines). There's on top of that code-style issues.

Considering all the above and current state of v2 while v3 is indev, this pr will probably not be merged, the benefits being too minor in this context.

An00nymushun:
I didn't know hrtime was in nanoseconds, but both are high precision, it's just that the performance one is multiplied so it's milliseconds plus fractions.

I don't get what you mean by code-style issues and where is the code for v3?

PR #572, opened 2021-11-07
Hungarian translation and bug fixes

Chri6s:
Translated by FrimeX#6701

No Google translate, only Oxford dictionary.


cyyynthia:
We are not accepting new translations at this time - for future reference translations should be done on https://i18n.powercord.dev

Note that it is a really bad practice to mix multiple unrelated changes in a single PR as it makes review difficult and invalidates the whole PR if a single thing is wrong.

Issue #567, opened 2021-10-24
Emote Utilities mass emote not working


Sh1mz:
**Describe the bug**
Emote Utilities' mass emote not working, with the error `An error occurred while executing the command: this.getCurrentUser is not a function.`

**To Reproduce**
Run the mass emote command

**Expected behavior**
Should "send all emotes containing the specified name"

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Powercord version: 1d2318b

**Additional context**
 - [✅] I checked if Discord and/or Powercord are up to date
- In console it shows:
`An error occurred while executing command massemote: TypeError: this.getCurrentUser is not a function
    at Object.executor (C:\Users\Shimz\powercord\src\Powercord\plugins\pc-emojiUtility\index.js:601:21)
    at Object.messages.sendMessage.oldSendMessage [as sendMessage] (C:\Users\Shimz\powercord\src\Powercord\plugins\pc-commands\monkeypatchMessages.js:25:30)`

FrimeX:
I looked into this and this all goes back to the `doImport()` function in the start. The function itself is just `await this.import()`. This functions params comes down to `(filter: any, functionName?: any)` and in the index.js of pc-emoteutility it only uses the filter param. I added the line `await this.import('getCurrentUser', 'getCurrentUser')` and I (Sadly i dont have nitro at the moment so i cant really test it) didnt see any exceptions so it must be an easy fix.


TL; DR: The function's parameters are used wrongly in doImport() and just needs to rewrite `await this.import([ 'getCurrentUser', 'getUser' ]);` to `await this.import('getCurrentUser', 'getCurrentUser')`.

PR #573, opened 2021-11-11
Fix for #567 : EmojiUtilities massemote not working

vinceh121:
I rewrote `import()` to support multi-filters with function renaming, as originally it supported filters with only a single function.


Chri6s (approved):
Seems legit.

vinceh121:
Fixed in b64b144959a922a41e4d15122b195354577bc582

PR #578, opened 2021-11-15
Fixed the main Settings Panel error.

Ciach0:
Removed the `format()` function from the `Messages.POWERCORD_SETTINGS_EXP_WEB_PLATFORM_DESC` string, because it causes the main Powercord Settings Panel not to render because of a ReferenceError.


cyyynthia:
The call to format is required to actually render the string as it has markdown in it (to render the bold Requires Restart).
image

PR #593, opened 2021-11-26
fix: ignore .DS_Store when loading modules

### LINKED ISSUE #571: opening src/Powercord/modules/ in Finder on macOS creates file, .DS_Store, causes powercord to crash on startup
**Describe the bug**
the finder app creates a file, .DS_Store in every folder you go into (yeah it's really not very smart), and powercord seems to think it's a module. it tries to load it, and obviously fails at doing so...preventing powercord from loading entirely and creating other (temporary) minor functional issues in discord

**To Reproduce**
simply opening up the modules folder is enough to do it

**Expected behavior**
PC should probably ignore non-modules and dotfiles

**Desktop (please complete the following information):**
 - OS: Mac
 - Powercord version: `Official 1d2318b v2` (up-to-date as of right now, not exaaaaaactly sure how to read these versions :p)

**Additional context**
 - [x] I checked if Discord and/or Powercord are up to date
 - [x] I made sure this is not related to an external plugin or a theme

katlyn:
closes powercord-org/powercord#571

### DIFF
--- a/src/Powercord/modules/index.js
+++ b/src/Powercord/modules/index.js
@@ -1,4 +1,4 @@
 module.exports = require('fs')
   .readdirSync(__dirname)
-  .filter(file => ['index.js', '.DS_Store'].includes(file))
+  .filter(file => !['index.js', '.DS_Store'].includes(file))
   .map(filename => require(`${__dirname}/${filename}`));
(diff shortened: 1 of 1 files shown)


cyyynthia:
This fix is incomplete and does not cover the 19 other places where this bug can occur. A more complete fix has been rolled out in b64b144959a922a41e4d15122b195354577bc582.

PR #599, opened 2021-12-22
Fixed search plugins divider

cafeed28:
😼 
image


GriefMoDz:
Thanks for the pull request. Unfortunately, we can't accept this as the styling of which you're trying remove is still actively used - we purposely hide the divider below the search bar to keep things looking reasonably tidy on the "Plugins" tab. For this reason I'll be closing this pull request.

PR #602, opened 2022-01-09
Better injector handling on Linux

oatmealine:
Currently, if a user is using a non-standard Discord install location, they will have to enter it each time they inject and uninject Powercord. This PR will fix this by creating an `.installdir` file and using that if possible.
Additionally, the script will fail if the user uses `doas` (and doesn't have a `sudo` shim or similar). It will now instead check for both, then fail gracefully if neither are found.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ settings/*.json
 config.json
 *injected.txt
 *.nogit.*
+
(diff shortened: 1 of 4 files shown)


asportnoy (commented):
This needs a rebase

PR #608, opened 2022-02-02
bump year in the license

Assassin-1234:
Bumped the year on the outdated license


cyyynthia:
This is an unnecessary PR.

PR #614, opened 2022-02-11
Fix Emoji Utility

TheRealClarity:
A bit dirty, but works. Had to use lazy loading due to the old way not finding "MessageContextMenu". This has the consequence of the Clone menu only appearing after the right click menu has been loaded once. Is there a way to fix that?

### DIFF
--- a/src/Powercord/plugins/pc-emojiUtility/index.js
+++ b/src/Powercord/plugins/pc-emojiUtility/index.js
@@ -243,7 +243,7 @@ module.exports = class EmojiUtility extends Plugin {
           }
 
           if (guild) {
-            if (!this.hasPermission(guild.id, Permissions.MANAGE_EMOJIS_AND_STICKERS)) {
+            if (!this.hasPermission(guild.id, Permissions.MANAGE_GUILD_EXPRESSIONS)) {
               return this.replyError(`Missing permissions to upload emotes in **${guild.name}**`);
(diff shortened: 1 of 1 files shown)


cyyynthia:
Thanks for the PR, the fix has been implemented in a different way in e398071d5d03526f377e7bce061bd7746aca755b with the inclusion of the context menu injection code as a globally available helper in bcc3edef1e4d7063439afbc1fd5fa91cb3340255, which does not require 2 right clicks to work.

TheRealClarity:
> Thanks for the PR, the fix has been implemented in a different way in [e398071](https://github.com/powercord-org/powercord/commit/e398071d5d03526f377e7bce061bd7746aca755b) with the inclusion of the context menu injection code as a globally available helper in [bcc3ede](https://github.com/powercord-org/powercord/commit/bcc3edef1e4d7063439afbc1fd5fa91cb3340255), which does not require 2 right clicks to work.

Hey @cyyynthia, thanks for looking at my PR! Discord broke more than injection though :(
Emoji url's are now .webp instead of .png, and Discord doesn't allow adding .webp images as emotes. To fix this, you need to change `https://cdn.discordapp.com/emojis/xyz.webp` to `https://cdn.discordapp.com/emojis/xyz.png` when cloning the emote. I did this with line 787 of my PR.

PR #633, opened 2022-04-25
Add commands for plugging stable, PTB, and development

### LINKED ISSUE #632: Allow stable/PTB injection even though it is not supported
The README explains the reason for Powercord being canary-only is that internal differences could cause non-canary versions to crash. However, canary also frequently experiences issues with crashes when Discord makes a change.

Allowing Powercord to be used on these versions would be beneficial as canary has worse performance. In addition, the other versions could allow users to continue to use Powercord when canary breaks due to a Discord changes.

I propose you add an option to inject to non-canary versions, maybe with a flag on the plug command. This could contain a warning that non-canary versions may experience issues and support will not be provided.

### LINKED ISSUE #647: Add Discord Development support
Discord Development is basically discord canary (The Version is exact the same I could'nt find any difference).
So why it isnt supported?

asportnoy:
This adds commands for installing to Stable or PTB. This can be done by appending a version to the command: `npm run <plug|unplug> [stable|ptb|canary|dev|development]`. A warning will be given to the user and they will have to confirm before the install continues. 
Screenshot on 2022-04-24 at 22 13 23

I was only able to test injection on macOS. I was able to find the windows paths [here](https://www.reddit.com/r/discordapp/comments/dhiovp/comment/f3ofvrl/?utm_source=share&utm_medium=web2x&context=3) but I had to guess the Linux paths based on context.

I updated the info in the README, but the wiki still needs to be updated. You might want to move the README info to the wiki instead of where it currently is, but that is up to you. The information in the Discord server probably needs to be updated as well.

Closes #632
Also closes #647

### DIFF
--- a/injectors/index.js
+++ b/injectors/index.js
@@ -3,9 +3,26 @@ require('./env_check')(); // Perform checks
 require('../polyfills'); // And then do stuff
 
 const { join } = require('path');
-const { readFile, writeFile } = require('fs').promises;
-const { BasicMessages } = require('./log');
+const { writeFile } = require('fs').promises;
+const readline = require('readline');
+const { AnsiEscapes, BasicMessages } = require('./log');
 const main = require('./main.js');
+
+async function promptYesNo (question) {
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout
+  });
+
+  return new Promise(resolve => {
+    rl.question(question, answer => {
+      rl.close();
+      resolve(
+        answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes'
+      );
+    });
+  });
+}
 
 let platformModule;
 try {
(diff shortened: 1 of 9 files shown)


Captain8771:
is there any reason discordDevelopment isnt included?

asportnoy:
> is there any reason discordDevelopment isnt included?

As far as I know, there's no advantage discordDevelooment has and it's not supposed to be used anyways

kekkodance:
> > is there any reason discordDevelopment isnt included?
> 
> As far as I know, there's no advantage discordDevelooment has and it's not supposed to be used anyways

it's not about "advantages". if a user chose that they needed powercord on dev (which would be basically the same as a canary instance) they would have to do the rudimentary method of changing a string in powercord's injector, which makes your PR useless. it's not hard to add a new "dev" instance in the choices.

asportnoy:
Alright. Let me know the install paths and the subdomain and I'll add it later today.

kekkodance:
the pathname for dev is just "DiscordDevelopment"

asportnoy:
And is it dev.discord.com? or development.discord.com?

kekkodance:
discord dev just points to canary, doesn't it? they offer the installer for dev here: https://discordapp.com/api/download/development?platform=win 
but they do not have another endpoint like ptb.discord.com / canary.discord.com etc.

CanadaHonk:
> discord dev just points to canary, doesn't it

Just to confirm, yes it just loads the normal canary webapp_endpoint (`https://canary.discord.com/app`).

asportnoy:
As with the other clients, I had to guess on the linux paths. But macOS and Windows should both be good.

injectors/index.js
```diff
@@ -52,3 +69,21 @@ try {
...
+  });
+
+  return new Promise((resolve, reject) => {
+    rl.question(question, answer => {
+      rl.close();
+      if (answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes') {
+        resolve(true);
+      } else {
+        resolve(false);
+      }
```
aetheryx: resolve(answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes')

injectors/linux.js
```diff
@@ -7,33 +7,62 @@ const { BasicMessages, AnsiEscapes } = require('./log');
...
+    `${homedir}/.local/bin/DiscordCanary` // https://github.com/powercord-org/powercord/pull/370
+  ]),
+  dev: Object.freeze([
+    '/usr/share/discord-development',
+    '/usr/lib64/discord-development',
+    '/opt/discord-development',
+    '/opt/DiscordDevelopment',
+    `${homedir}/.local/bin/DiscordDevelopment`
+  ])
+};
```
aetheryx: freeze the entire object (`const KnownLinuxPaths = Object.freeze({ stable: [ ... ], ... })`) instead of each of the arrays

injectors/linux.js
```diff
@@ -7,33 +7,62 @@ const { BasicMessages, AnsiEscapes } = require('./log');
...
       const readlineInterface = readline.createInterface({
         input: process.stdin,
         output: process.stdout
       });
 
       const askPath = () => new Promise(resolve => readlineInterface.question('> ', resolve));
-      console.log(`${AnsiEscapes.YELLOW}Failed to locate Discord Canary installation folder.${AnsiEscapes.RESET}`, '\n');
-      console.log('Please provide the path of your Discord Canary installation folder');
+      console.log(`${AnsiEscapes.YELLOW}Failed to locate Discord${platform  == 'canary' ? ' Canary' : ''} installation folder.${AnsiEscapes.RESET}`, '\n');
+      console.log(`Please provide the path of your Discord${platform  == 'canary' ? ' Canary' : ''} installation folder`);
```
aetheryx: the platform should always be printed, just print something like `Failed to locate Discord (platform: ${platform})`

aetheryx (changes requested):

injectors/index.js
```diff
@@ -52,3 +69,17 @@ try {
...
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  return new Promise((resolve, reject) => {
+    rl.question(question, answer => {
+      rl.close();
+      resolve(answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes')
+    })
```
aetheryx: missing semicolons

injectors/linux.js
```diff
@@ -2,38 +2,67 @@ const { join } = require('path');
...
+    `${homedir}/.local/bin/DiscordDevelopment`
+  ]
+});
 
+const ProcessRegex = {
+  stable: /discord$/i,
+  ptb: /discord-?ptb$/i,
+  canary: /discord-?canary$/i,
+  dev: /discord-?development$/i
+}
```
aetheryx: missing semicolon

src/patcher.js
```diff
@@ -73,15 +73,21 @@ electron.app.once('ready', () => {
...
       // It should get restored to _powercord url later
-      done({ redirectURL: 'https://canary.discord.com/app' });
+      done({ redirectURL: `${domain}/app` });
     } else {
       done({});
     }
+  } else {
+    done({});
+  }
   });
```
aetheryx: fix indentation

aetheryx (changes requested):

katlyn (approved):
All changes work on linux clients after cursory testing.

asportnoy:
Rebased

PR #636, opened 2022-04-27
Merge PowercordPluginDownloader into pc-moduleManager

LandenStephenss:
merged PowercordPluginDownloader into pc-moduleManager, also fixed a bug with pc-tags adding a tag that has no content.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   },
   "devDependencies": {
     "eslint": "^7.8.1",
-    "eslint-plugin-react": "^7.20.6",
+    "eslint-plugin-react": "^7.29.4",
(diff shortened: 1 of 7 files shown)


LandenStephenss:
@aetheryx, mind review/testing this for me? It worked with the 'Install Theme' button but for some reason it wouldn't show up again after doing it once or twice. Not sure what caused that (i am also currently downloading something with my slow internet so that might be affecting it since it checks for powercord_manifest and manifest).

Edit: It does work after a bit of testing on my end. I am going to add a "cache" so that plugins and themes don't have to be "recognized" (i.e fetch manifest)

asportnoy:
This should be closed

PR #644, opened 2022-05-19
fix: don't crash on command autocomplete

### LINKED ISSUE #643: pc-commands crash
**Describe the bug**
When you type `.` it's supposed to show the command but it won't work anymore, it just make you crash.

**To Reproduce**
- Plug the last version of powercord
- Relaunch
- Type `.`

**Expected behavior**
Show the commands with TITLE
But instead of doing this, `TypeError: Class constructor Title cannot be invoked without 'new'` at [React Title Return](https://github.com/powercord-org/powercord/blob/v2/src/Powercord/plugins/pc-commands/injectAutocomplete.js#L12)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Powercord version: V2 (02bd12f)

**Additional context**
 - [x] I checked if Discord and/or Powercord are up to date
 - [x] I made sure this is not related to an external plugin or a theme
 
When I replace it to a simple text or `[ 'Powercord ', title ]` it work perfectly.

katlyn:
This commit fixes #643 by using the vanilla Autocomplete.Title component. After testing, there is no regression that I could find from using the unpatched component rather than custom patched version.

### DIFF
--- a/src/Powercord/plugins/pc-commands/components/Title.jsx
+++ b/src/Powercord/plugins/pc-commands/components/Title.jsx
@@ -1,5 +1,19 @@
 const { React, getModuleByDisplayName } = require('powercord/webpack');
+const { AsyncComponent } = require('powercord/components');
 
-const Autocomplete = getModuleByDisplayName('Autocomplete', false);
+module.exports = AsyncComponent.from(
+  (async () => {
+    const Autocomplete = await getModuleByDisplayName('Autocomplete');
 
-module.exports = Autocomplete.Title;
+    return (props) => {
+      const res = Autocomplete?.Title?.(props);
+      const EmptyContainer = <div style={{ padding: '4px' }}/>;
+
+      if (Array.isArray(props?.title) && !props.title[0]) {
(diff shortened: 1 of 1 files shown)


HideakiAtsuyo (approved):
Work as expected, not sure to know why they wanted to patch 🤷

GriefMoDz:
Call it pointless, but the sole purpose of using a custom autocomplete `Title` component is simply to take into account empty autocomplete titles from which the native component doesn't support.

The following figures show how this is applied through the use of the `lmgtfy` command with the "Display Autocompletes" setting enabled.

1. Using the native component (i.e. displays a large empty space where the title would normally be):
Figure 1

2. Using the custom component (i.e. removes the large empty space where the title would normally be):
Figure 2

How I would go about fixing this instead, is by wrapping the custom component around an `AsyncComponent` and making it fallback to an empty container if the `Title` component isn't found or returns an empty result.

It should end up looking something like [this](https://haste.powercord.dev/rezoquwowu.js) and will additionally help mitigate such crashes from arising again.

PR #646, opened 2022-05-21
Fix: injectAutocomplete.js

cKnei:
Fix for developers who forget to add a command key to their command so instead of crashing, it just logs the error.

### DIFF
--- a/src/Powercord/plugins/pc-commands/injectAutocomplete.js
+++ b/src/Powercord/plugins/pc-commands/injectAutocomplete.js
@@ -31,7 +31,12 @@ module.exports = async function injectAutocomplete () {
   }
 
   function getMatchingCommand (c) {
-    return [ c.command.toLowerCase(), ...(c.aliases?.map((alias) => alias.toLowerCase()) || []) ];
+    try {
+      return [ c.command.toLowerCase(), ...(c.aliases?.map((alias) => alias.toLowerCase()) || []) ];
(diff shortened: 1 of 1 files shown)


src/Powercord/plugins/pc-commands/injectAutocomplete.js
```diff
@@ -31,7 +31,12 @@ module.exports = async function injectAutocomplete () {
   }
 
   function getMatchingCommand (c) {
-    return [ c.command.toLowerCase(), ...(c.aliases?.map((alias) => alias.toLowerCase()) || []) ];
+    try {
+      return [ c.command.toLowerCase(), ...(c.aliases?.map((alias) => alias.toLowerCase()) || []) ];
+    } catch (e) {
+      console.error('%c[Powercord:Plugin:pc-commands]', 'color: #7289da', e);
```
katlyn: Using [this.error(...)](https://github.com/powercord-org/powercord/blob/v2/src/fake_node_modules/powercord/entities/Plugin.js#L106) would likely be preferred, if you can get it into this file from index.js

cKnei: Just made a new commit that uses the suggestion, I do have a have a question on how well, custom error messages go

cKnei:
To whoever reviews this I need some input on the error message(line 38?) that gets sent, and what should change

PR #654, opened 2022-06-10
PowercordPluginDownloader integration

oillicker:
This pull request integrates PowercordPluginDownloader. It is working 100% (context menus and the message popover for both themes and plugins). Feel free to test on your own, also ran `npm run lint` to fix a few of those things.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   },
   "devDependencies": {
     "eslint": "^7.8.1",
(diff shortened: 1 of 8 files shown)


oillicker:
Should pc-moduleManager check if `PowercordPluginDownloader` or the theme downloader is installed and disable it if it is? 
This would mostly be done just to avoid conflicts between having the two run at once

PR #655, opened 2022-06-14
fix: emoji cloning

12944qwerty:
Fixes the emoji cloning problem.


asportnoy (changes requested):
Please run the linter. `npm run lint -- --fix`

PR #657, opened 2022-06-18
Updated changelog

oillicker:
This pull request updates the changelog for powercord.
The changelog file is updated to be an array rather than an object so that old changelogs are kept for whatever reason we may want them, and the first one is fetched from the file.
image

### DIFF
--- a/changelogs.json
+++ b/changelogs.json
@@ -1,52 +1,88 @@
-{
-  "id": "updates-26-04-20",
-  "date": "2019-11-17",
-  "contents": [
-    {
-      "type": "HEADER",
-      "color": "GREEN",
-      "text": "Update time!",
-      "noMargin": true
-    },
-    {
-      "type": "CONTENTS",
-      "text": "It's been months since last time, so here we go with updates: Themes tab in settings is no longer useless. It's not the complete theme UI yet, this is still under construction, but you can at least apply some CSS more easily!",
-      "list": [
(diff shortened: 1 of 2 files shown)


changelogs.json
```diff
@@ -1,52 +1,88 @@
...
+      },
+      {
+        "type": "HEADER",
+        "color": "BLURPLE",
+        "text": "Other changes and fixes"
+      },
+      {
+        "type": "CONTENTS",
+        "list": [
+          "You can now inject into other Discord clients (Stable and PTB), although it is **__not recommended and unsupported__**, [more info](https://github.com/powercord-org/powercord#how-can-i-install-powercord-on-stable-or-ptb).",
```
asportnoy:
```suggestion
          "You can now inject into other Discord clients (Stable and PTB), although it is **__not recommended and unsupported__**. [More info](https://github.com/powercord-org/powercord#how-can-i-install-powercord-on-stable-or-ptb)."
```

asportnoy (commented):
What's the reasoning behind linking to the issues? It seems unnecessary to me. And if you're going to link to issues, the other 3 fixes should also have links for consistency

oillicker:
> What's the reasoning behind linking to the issues? It seems unnecessary to me. And if you're going to link to issues, the other 3 fixes should also have links for consistency

Firstly they're pull requests not issues, but I linked the PRs I could find.

asportnoy:
Still, I don't think there's any point in including them.

PR #658, opened 2022-06-21
pc-moduleManager: Add install/uninstall commands

asportnoy:
- Move code to get repo info to own file and refactor link context menu accordingly
- Add command to install plugin or theme. Takes either a GitHub link or `username/reponame` as an argument.
- Add command to uninstall plugin or theme.

### DIFF
--- a/src/Powercord/plugins/pc-moduleManager/commands/disable.js
+++ b/src/Powercord/plugins/pc-moduleManager/commands/disable.js
@@ -1,5 +1,5 @@
 const { CORE_PLUGINS } = require('powercord/constants');
-const { resp } = require('../util/resp');
+const { resp } = require('../util');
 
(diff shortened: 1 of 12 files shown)


CanadaHonk:
Shouldn't there be at least some warning/confirmation about danger/trust on install? (As this makes essentially RCE as a feature in ~three clicks with no warning?)

asportnoy:
Good point. I can add a modal to confirm install and I'll add it for uninstall too.

CanadaHonk:
Yeah probably want to check user actually wants to uninstall and not by mistake (?) or something.

asportnoy:
Modal confirmation is done.

asportnoy:
Closing due to EOL

PR #659, opened 2022-06-24
feature: react developer mode

12944qwerty:
Powercord Developer Mode settings. Currently on the react developer tools that show Components and Profiler in inspect.

I'm gonna add a developer SDK


asportnoy (changes requested):
Pressing "restart" while devtools are open gives this error.
Screenshot on 2022-06-24 at 08 57 18

12944qwerty:
Ok, so. I noticed that pc-sdk existed so I'm just gonna continue off of what Cynthia had started. 
It was extremely broken so I tried fixing as much as I could. 
I also added a powercord Labs section in settings so that you can enable and disable experiments. I think only the sandbox dev kit will work, not sure though 
However, the SDK is always on right now (even though the switchitem doesn't show it as such). This is just for me testing ~~and definitely not because I forgot to turn it off~~

My ToDo for the SDK right now is as follows, and I will add more later on
- [ ] Fix any existing features Cynthia added
- [ ] Integrate [PowercordModuleFinder](https://github.com/VenPlugs/PowercordModuleFinder)
- [ ] Integrate [F8Fix](https://github.com/doggybootsy/F8Fix)
- [ ] Maybe, just maybe, add a lab or fix labs (not sure yet)

PR #668, opened 2022-07-17
coremods/badges: fix

CanadaHonk:
Fixes user badges injection, no user badges are added currently/without this PR. Also fixes badges (including guild badge) not appearing.


asportnoy (approved):
Can confirm both fixes are working.

12944qwerty (approved):

PR #669, opened 2022-07-19
pc-updater: fix when not on branch

asportnoy:
Properly get revision when not on a branch (previously would fail and just display `???`)
Allow branch and upstream to be null
Screenshot on 2022-07-19 at 13 13 16
Screenshot on 2022-07-19 at 13 13 28


asportnoy:
Closing due to EOL

PR #670, opened 2022-07-20
Qwerty theme toggling

aetheryx:


src/Powercord/plugins/pc-moduleManager/commands/enable.js
```diff
@@ -1,46 +1,57 @@
...
-            (${args[0]})`;
-      }
-    } else {
-      result = `->> ERROR: Tried to enable a non-installed plugin!
-          (${args[0]})`;
+  description: 'Enable a plugin/theme',
+  usage: '{c} [ plugin/theme ID ]',
+  executor([ id ]) {
+    const isPlugin = powercord.pluginManager.plugins.has(args[0]);
+    const isTheme = powercord.styleManager.themes.has(args[0]);
```
asportnoy:
```suggestion
    const isPlugin = powercord.pluginManager.plugins.has(id);
    const isTheme = powercord.styleManager.themes.has(id);
```

asportnoy (changes requested):
Enable command doesn't work, see suggestion for fix

src/Powercord/plugins/pc-moduleManager/commands/disable.js
```diff
@@ -1,51 +1,72 @@
...
+
+const resp = (success, description) => ({
+  send: false,
+  result: {
+    type: 'rich',
+    color: success ? 0x1bbb1b : 0xdd2d2d,
+    title: success ? 'Success' : 'Error',
+    description,
+  },
+});
```
asportnoy: I think it would be better to move this to the util folder (in the plugin)

src/Powercord/plugins/pc-moduleManager/commands/enable.js
```diff
@@ -1,46 +1,57 @@
+const { resp } = require('./disable');
```
asportnoy: If you move the function to util make sure to update this

asportnoy (approved):
Code looks good, all features work as expected

12944qwerty (approved):
Works :D

PR #671, opened 2022-07-21
Fix the updater freezing the client

lleyton:
This PR fixes the updater freezing the client when checking for updates. See Discord discussion for more info.


asportnoy (approved):
Looks good to me! Fix works great, I can update everything concurrently without Discord slowing down at all!

PR #672, opened 2022-07-22
Replace more instances of child_process.exec to use IPC

lleyton:
Follow up to #671

P.S. My Git history is a bit messed up, so you might want to squash-merge this one


asportnoy (approved):
Haven't tested it but code looks fine.

asportnoy (approved):
Tested too now and still works fine

